PR #1, opened 2023-12-13
Activations on disk

lucyfarnik:


sae_training/activations_store.py
jbloomAus, line 1: seems like the use cached activation stuff could be better abstracted (eg: a set of private methods).
create data loader now can be false, which I can see why you'd want it but we haven't handled when it's false (an error or check might be good)

sae_training/cache_activations_runner.py
jbloomAus, line 1: looks good

sae_training/config.py
jbloomAus, line 1: I wouldn't have done it this way. Seems like it ties two things together that don't need to be. I think managing this via a clean interface that lm runner uses when using cached activations would be best.

sae_training/utils.py
jbloomAus, line 1: why not have shuffle activations sit in the activation store module or as a method on the class?

PR #9, opened 2024-02-25
chore: removing .DS_Store files

chanind:
It looks like this was added to .gitignore in 1ec8f97330cc605d003757b9c8d6edb8040119bc, but these files weren't removed


jbloomAus:
thanks!

PR #11, opened 2024-02-26
Fixed edge case in activation cache shuffling

lucyfarnik:
There was a bug where you could pass `buffer_idx_range=(0, 1)` into `shuffle_activations_pairwise`, it wouldn't raise an exception like it was supposed to and then get stuck in an infinite loop trying to find two non-identical indices within that range.

This was particularly likely to happen during the final shuffling when caching activations, so I've added an if statement there to make sure final pairwise shuffling doesn't happen if there's only 1 cache file.


jbloomAus:
Thanks!

PR #13, opened 2024-02-26
chore: using poetry for dependency management

chanind:
This PR sets up [Poetry](https://python-poetry.org/) for dependency management.

In the process of adding this, this PR does the following:
- removes `pylint` from `requirements.txt` and replaces it with `flake8`, since all the linting in CI uses flake8.
- adds `isort` since it's referenced in the `.vscode` config, but seemed to be absent from the actual requirements
- updates the Github Actions config to install via Poetry, and also check `black` and `isort` formatting
- updates the `Makefile` to use Poetry, and fills in the `make format` and `make check-format` commands
- runs `black` and `isort`, which resulted in a lot of reformatting
- Updates the `README` to reference Poetry

This PR leaves the following out:
- Leaves `requirements.txt` in case users are already relying on this to install dependencies. This can be removed if needed (and probably should be at some point)
- Does not add the `poetry.lock` file to this repo. Since this will be eventually published as a library, it's best not to include the lockfile in the repo IMO.
- Does not move the main `sae_training` and `sae_analysis` to a common subfolder. This will need to be done before publishing this as a PyPI module.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -42,18 +42,20 @@ jobs:
         uses: actions/setup-python@v4
         with:
           python-version: ${{ matrix.python-version }}
-          cache: 'pip'
+      - name: Install Poetry
+        uses: snok/install-poetry@v1
       - name: Install dependencies
-        run: |
-          python -m pip install --upgrade pip
-          pip install flake8 pytest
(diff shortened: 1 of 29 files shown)


jbloomAus:
Thanks, this is really great work!

PR #16, opened 2024-02-27
chore: enable full flake8 default rules list

### LINKED ISSUE #14: Use flake8 `extend-select` instead of `select`
it looks like flake8 linting is configured using `select = E9, F63, F7, F82`, but this effectively overrides all the flake8 default linting rules and runs ONLY those 4 rules. Most likely, the intention was to use `extend-select = E9, F63, F7, F82`, so that all the default flake8 rules still apply, but those 4 rules apply as well. More info on extend-select vs select: https://flake8.pycqa.org/en/latest/user/options.html#cmdoption-flake8-extend-select

It also looks like the Github action for flake8 is manually setting options as well, and running twice. Why is this the case rather than simply running `flake8 .` in the action, using the settings specified in `.flake8`?

chanind:
This PR enables the full set of flake8 default rules by using `extend-select` and `extend-ignore` instead of `select` and `ignore`.

 In addition, this PR makes the following notable changes:
- Simplifies the Github action to just run `flake8 .` so it will use the same settings as the flake8 on local
- Ignores the `geom_median` dir, since that's its own project, and its unclear if that should be linted by this
- ignores `E402` in `dashboard_runner.py` and `generate_dashboards.py` due to those files having `sys.path.append()` calls above imports. It's not clear to me why those calls are there, so I just left a TODO and ignored the error
- removes `# noqa` statements that didn't seem necessary
- adds `r` to regex strings, which I think was probably the intention anyway

closes #14

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -1,7 +1,8 @@
 [flake8]
-ignore = E203, E266, E501, W503
+extend-ignore = E203, E266, E501, W503, E721, F722, E731
 max-line-length = 79
-max-complexity = 10
-select = E9, F63, F7, F82
+max-complexity = 25
+extend-select = E9, F63, F7, F82
 show-source = true
 statistics = true
+exclude = ./sae_training/geom_median/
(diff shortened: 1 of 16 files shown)


jbloomAus:
@chanind Thanks so much!

PR #17, opened 2024-02-28
SAE Group for sweeps PR

Benw8888:
I implemented the SAEGroup class, which is basically a list of autoencoders, and which automatically initializes autoencoders based on which parameters of the config are passed as lists. Then I modified the training code to handle using SAEGroups.


jbloomAus:
@chanind  I'm having trouble reproducing the isort check fail locally which is somewhat frustrating. Don't suppose you could take a look and fix it? Would prefer to turn off the isort check if this delays us merging this for more than a day.

chanind:
weird, I'm able to reproduce these errors locally. The commands I'm running are:

To display the error (same as what CI is running):
```
poetry run isort . --check-only --diff
```

To fix the errors:
```
poetry run isort .
```

What are these commands outputting when you run them?

Let me see if I can figure out how to make a PR to this fork's PR...

chanind:
I opened up a PR to @Benw8888's branch after running isort here: https://github.com/Benw8888/mats_sae_training/pulls/1. Merging this PR should fix the isort issue. I'm still confused why running `isort` isn't working for you both locally 🤔.

jbloomAus:
Thanks @Benw8888 @chanind !

PR #20, opened 2024-03-06
chore: adding more tests to ActivationsStore + light refactoring

chanind:
This PR contains a few new tests for ActivationsStore (pretty modest though), and some light refactoring of the test setup code and ActivationsStore. Specifically, this PR changes the following:

- Adds `tests/helpers.py` where generic test utils can be put
- Moves the set of the `LanguageModelSAERunnerConfig` to `build_sae_config` helper in `tests/helpers.py`. This helper defines a default config, but allows any overrides to be passed in as kwargs. This is trying to be a lightweight version of the test factory pattern, like in https://github.com/FactoryBoy/factory_boy. This lets tests just override only the keys that are important for the test, and makes it more clear what part of the config is relevant for the test.
- Fixes a minor bug in `ActivationsStore` where sometimes the BOS token is added twice at the start of a batch
- Simplifies some duplicated code in `ActivationsStore`, and breaks out some of the token selecting logic into a private function for easier testing.
- Allows passing a `dataset` as an explicit param to `ActivationStore` constructor, to allow for testing with a small dataset that the individual tests can control and make assertions against.

Happy to change or remove any of this if it's not what's desired!


jbloomAus:
Hey David, looks great! I'd love to experiment with doing benchmark runs before/after changes. Do you think you might be able to do like a tiny-stories training run with / without these changes so we can see the difference? Happy to help tune hyperparameters.

chanind:
Yeah, definitely, I'll give that a try. Maybe we can get a snapshot test or something similar set up on a small dataset which can run automatically in CI and give confidence that the overall system is working. I see there's benchmark tests that have the shell of something like that, but are missing assertions that would give confidence in the overall results. I'll work on setting up something like that and reach out when there's something we can look over

jbloomAus:
@chanind How's this going? Just checking in.

chanind:
Apologies for being MIA on this! I've just been busy with other projects the past few days, but I should be able make progress on this this week.

PR #40, opened 2024-03-22
Refactor train SAE and adding unit tests

chanind:
This PR refactors the `train_sae_on_language_model` function, breaking it into smaller sub-functions which can be tested individually.

This PR makes the following changes to the public API:
- The `train_sae_on_language_model` is deprecated but still works as before
- Instead, `train_sae_group_on_language_model` is the preferred way to train - the `train_sae_on_language_model` is confusingly named since it actually runs on a `SAEGroup`, not a `SparseAutoencoder`.

Internally, this PR makes the following changes:
- The state that needs to be tracked for each SAE being trained is moved to a `SAETrainContext` object which is passed around internally. This just collects `act_freq_scores`, `n_forward_passes_since_fired`, `n_frac_active_tokens`, `optimizer`, and `scheduler` into a single struct
- functions that are just meant to be used internally in training are prefixed with `_`. I'm not sure if this is a good convention to follow - is this overkill?
- the `train_sae_on_language_model` function is broken into the following sub-functions: `_init_sae_group_b_decs`, `_build_train_context`, `_train_step`, `_build_train_step_log_dict`, `_save_checkpoint`, and `_log_feature_sparsity`. The most important and most thoroughly tested of these is `_train_step`, which runs a single training step for 1 autoencoder.
- `scheduler.step()` is moved below `optimizer.step()` - Pytorch complains if this is not the case, and likely stepping the scheduler before the optimizer was a mistake


sae_training/sparse_autoencoder.py
```diff
@@ -138,7 +147,14 @@ def forward(self, x: torch.Tensor, dead_neuron_mask: torch.Tensor | None = None)
         l1_loss = self.l1_coefficient * sparsity
         loss = mse_loss + l1_loss + mse_loss_ghost_resid
 
-        return sae_out, feature_acts, loss, mse_loss, l1_loss, mse_loss_ghost_resid
+        return ForwardOutput(
```
chanind, line 150: This should be equivalent to returning a plain tuple, since `NamedTuple` extends tuple. This is just so we can also access the fields by name if we want to make sure we don't accidentally mix up which loss goes in which position when reading or writing from this tuple

sae_training/train_sae_on_language_model.py
```diff
@@ -95,240 +258,195 @@ def train_sae_on_language_model(
...
-                            f"metrics/l0{suffix}": l0.item(),
-                            # sparsity
-                            f"sparsity/mean_passes_since_fired{suffix}": n_forward_passes_since_fired[
-                                i
-                            ]
-                            .mean()
-                            .item(),
-                            f"sparsity/dead_features{suffix}": ghost_grad_neuron_mask.sum().item(),
-                            f"details/current_learning_rate{suffix}": current_learning_rate,
-                            "details/n_training_tokens": n_training_tokens,
```
chanind, line 212: This key doesn't get the `{suffix}` added - was this intentional?

sae_training/train_sae_on_language_model.py
```diff
@@ -95,240 +258,195 @@ def train_sae_on_language_model(
...
 
-    # Generate the suffix by iterating over the keys that have list values in cfg
-    suffix = "".join(
-        f"_{key_mapping.get(key, key)}{getattr(hyperparams, key, '')}"
-        for key, value in vars(cfg).items()
-        if isinstance(value, list)
-    )
-    return suffix
+def _log_feature_sparsity(
+    feature_sparsity: torch.Tensor, eps: float = 1e-10
```
chanind, line 450: A lot of these epsilons are too small to be usable in fp16, Is it OK to increase this to ~1e-7 instead so fp16 works as well?

sae_training/train_sae_on_language_model.py
```diff
@@ -95,240 +258,195 @@ def train_sae_on_language_model(
...
+        sae_in,
+        ghost_grad_neuron_mask,
+    )
+    did_fire = (feature_acts > 0).float().sum(-2) > 0
+    ctx.n_forward_passes_since_fired += 1
+    ctx.n_forward_passes_since_fired[did_fire] = 0
+
+    with torch.no_grad():
+        # Calculate the sparsities, and add it to a list, calculate sparsity metrics
+        ctx.act_freq_scores += (feature_acts.abs() > 0).float().sum(0)
```
chanind, line 343: It seems like there's some inconsistenties in how sparsity is calculated in different parts of the code. Here, we use `acts.abs() > 0`, while for `did_fire`, we just use `acts > 0`. Also, we do `sum(0)` here, but `sum(-2)` for `did_fire`. I think these are all equivalent, since the acts have 2 dimensions, (so 0 == -2), and there's a ReLU in the autoencoder, so `feature_acts.abs() > 0` is the same as `feature_acts > 0` probably. Still, this feels like it could cause bugs in the future if we change the activation function of the autoencoder in the future. We should probably calculate this stuff the same way in both places.

sae_training/evals.py
```diff
@@ -3,11 +3,11 @@
 
 import pandas as pd
 import torch
-import wandb
 from tqdm import tqdm
 from transformer_lens import HookedTransformer
 from transformer_lens.utils import get_act_name
 
+import wandb
```
chanind, line 10: I think these import orderings are from isort, I didn't intentionally change these

jbloomAus: no worries

sae_training/sparse_autoencoder.py
jbloomAus: Love it!

sae_training/train_sae_on_language_model.py
jbloomAus: good catch. if this is tested, then setting them all to the same method + getting identical results on a small run seems like a good way to verify. Thanks for raising this.

sae_training/train_sae_on_language_model.py
jbloomAus: good catch. seems fine to me.

jbloomAus:
Thanks. 
I really like this PR:
- Really clear changes
- You flagged stuff for me to check which I valued. 
- You increased test coverage a bunch :D

PR #55, opened 2024-03-27
adding a contribution guide to docs

chanind:
This PR adds a contribution guide page to the docs, outlining how to get set up for development and what to run to make sure your code is up to par before opening a PR. This PR also adds some more makefile helpers to facilitate easier checking of code by contributors.

Some things to note:
- I didn't include anything about autodocs or docstring comments in the guide, since I don't think mkdocs supports autodoc. If we have a preferred docstring style we should consider adding a linting rule for it.
- I didn't add a demo `.vscode/settings.json` config, in case we change to use `ruff` for litning/formatting, as that would require a different vscode setup. If we want to stick with what we have, I'll add that to the docs too.

The rendered docs page is shown below
127 0 0 1_8000_about_contributing_

### DIFF
--- a/.github/workflows/deploy_docs.yml
+++ b/.github/workflows/deploy_docs.yml
@@ -36,6 +36,7 @@ jobs:
         run: poetry install --no-root
 
       - name: Deploy website
+        if: github.event_name == 'push' && github.ref == 'refs/heads/main'
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         run: poetry run mkdocs gh-deploy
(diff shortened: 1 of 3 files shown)


.github/workflows/deploy_docs.yml
```diff
@@ -36,6 +36,7 @@ jobs:
         run: poetry install --no-root
 
       - name: Deploy website
+        if: github.event_name == 'push' && github.ref == 'refs/heads/main'
```
chanind, line 39: This action was trying to deploy docs in the PR, which isn't ideal. This just makes sure we don't do the actual docs deploy step unless this is a real push to main.

jbloomAus:
Thanks David!

PR #56, opened 2024-03-27
minor refactoring to SAE and adding tests

chanind:
This PR is a minor refactoring of the SAE class, with added tests. The refactoring covers the following:
- moved the MSE with norm calculation into its own function, as it was used in 2 places
- rewrote the MSE with norm calculation to use torch built-in `mse_loss()` and `norm()`, but added tests to ensure this gives identical result. It reduced the amount of code to do it this way, and torch can be more efficient with these built-in functions.
- moved ghost grad calculation into a `calculate_ghost_grad_loss()` function to logically separate it from the rest of the forward pass
- removed unused  `initialize_b_dec_*` functions, as these are handled in the runner currently

Otherwise, this PR just adds test coverage. I also combined some of the save/load tests into a single parametrized test.

If any of the above changes don't feel right, I'm happy to revert any of them.


tests/unit/test_sparse_autoencoder.py
```diff
@@ -97,15 +66,10 @@ def sparse_autoencoder(cfg: Any):
 
 
 @pytest.fixture
-def model(cfg: SimpleNamespace):
+def model(cfg: LanguageModelSAERunnerConfig):
     return HookedTransformer.from_pretrained(cfg.model_name, device="cpu")
 
 
-@pytest.fixture
-def activation_store(cfg: SimpleNamespace, model: HookedTransformer):
```
chanind, line 105: This fixture didn't appear to be used by the tests

jbloomAus:
Thanks!

PR #57, opened 2024-03-27
Adding tests to get_scheduler

chanind:
This PR adds tests to the `get_scheduler()` function, and improves type hinting by listing all optional arguments instead of relying on `kwargs` so editors can auto-complete the available options.

In the process of writing these tests, I found we have an inconsistency with our `cosineannealingwarmup` which we implement manually, and  `cosineannealing`, which is implemented by pytorch. Specifically, what we're calling `lr_end` is not the same as `eta_min` if `lr_end != 0.0`. We need to either multiply `lr_end` by `lr` when passing to `eta_min` in `cosineannealing`, or divide `lr_end` by `lr` when using our `cosineannealingwarmup` function. This is likely not a big deal since I don't think we use `cosineannealingwarmup` by default, but we should be careful with this when using cosine annealing schedulers as the result may not be as expected.

### DIFF
--- a/tests/unit/test_optim.py
+++ b/tests/unit/test_optim.py
@@ -1,0 +1,140 @@
+import pytest
+import torch
+from torch.optim import Adam
+from torch.optim.lr_scheduler import (
+    CosineAnnealingLR,
+    CosineAnnealingWarmRestarts,
+    LRScheduler,
+)
+
+from sae_training.optim import get_scheduler
+
+
+@pytest.fixture
+def optimizer():
+    return Adam([torch.tensor(1.0)], lr=0.1)
+
+
+def step_times(num: int, optimizer: Adam, scheduler: LRScheduler):
+    for _ in range(num):
+        step(optimizer, scheduler)
+
+
+def step(optimizer: Adam, scheduler: LRScheduler):
+    optimizer.step()
+    scheduler.step()
+
+
+@pytest.mark.parametrize(
+    "scheduler_name",
+    [
+        "linearwarmupdecay",
+        "cosineannealing",
+        "cosineannealingwarmup",
+        "cosineannealingwarmrestarts",
+    ],
+)
+def test_get_scheduler_requires_training_steps(scheduler_name: str, optimizer: Adam):
+    with pytest.raises(AssertionError, match="training_steps must be provided"):
+        get_scheduler(scheduler_name, optimizer, 10)
+
+
+def test_get_scheduler_errors_on_uknown_scheduler(optimizer: Adam):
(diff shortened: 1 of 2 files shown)


tests/unit/test_optim.py
```diff
@@ -0,0 +1,140 @@
...
+    assert isinstance(scheduler, CosineAnnealingLR)
+    assert scheduler.T_max == 4
+    assert scheduler.eta_min == 0.05
+
+
+def test_get_scheduler_cosineannealingwarmup():
+    # NOTE: if the lr_end is not 0.0, this test will not pass.
+    # If eta_min = lr_end * lr, then the test will pass.
+    # We should be careful about the difference between our lr_end and proper eta_min.
+    lr_end = 0.0
```
chanind, line 100: changing this `lr_end` to something > 0 will highlight the bug where our cosine annealing and torch's cosine annealing define `eta_min` differently.

jbloomAus:
Great stuff. Will review properly soon. If you want to add a scheduler that
linearly approaches 0 in the final 20% of steps this would be useful as
Anthropic recommended it in their last update.

chanind:
Would it just be a constant scheduler with 20% linear decay?

One thing we could do with the `get_scheduler()` function is add a `linear_decay_steps` param similar to `warm_up_steps` param, and use [lr_scheduler.SequentialLR](https://pytorch.org/docs/stable/generated/torch.optim.lr_scheduler.SequentialLR.html#torch.optim.lr_scheduler.SequentialLR) to apply an optional linear warmup, followed by whatever the main scheduler is, followed by an optional linear decay. This would mean that any scheduler could have a linear warmup/decay using the sequentialLR to just apply a linear warmup/decay in series. That would simplify the function quite a bit as well.

jbloomAus:
> Would it just be a constant scheduler with 20% linear decay?
> 
> One thing we could do with the `get_scheduler()` function is add a `linear_decay_steps` param similar to `warm_up_steps` param, and use [lr_scheduler.SequentialLR](https://pytorch.org/docs/stable/generated/torch.optim.lr_scheduler.SequentialLR.html#torch.optim.lr_scheduler.SequentialLR) to apply an optional linear warmup, followed by whatever the main scheduler is, followed by an optional linear decay. This would mean that any scheduler could have a linear warmup/decay using the sequentialLR to just apply a linear warmup/decay in series. That would simplify the function quite a bit as well.

this sounds good! In general being able to compose the scheduler in various ways seems valuable.

jbloomAus:
I'm going to accept this and we can do the following improvements in another PR

PR #58, opened 2024-03-28
Make prepend BOS optional: Default True

canrager:


jbloomAus:
Looks great! Soons as the tests pass I will accept it :)

PR #60, opened 2024-03-29
fixing config typing

### LINKED ISSUE #44: SAEGroup config typing is incorrect
Currently, SAEGroup is configured using the same config object as normal SAE objects, but the typings are incorrect since SAEGroups allow lists for each item in the config which turn into a SAE config for every permutation of values in lists in the config, while the config for SAEs do not. This means that configuring a SAEGroup currently will report typing errors.

I can envision two possible solutions:

### Solution 1: Create a dedicated `LanguageModelSAEGroupRunnerConfig` class
One solution is for SAEGroup and SparseAutencoder classes to each have their own dedicated config class to address this issue, likely with similar fields but with SAEGroup allowing lists of items. This means being less DRY (don't repeat yourself) in the config code, but I feel it's justified in this case for the simplicity and clarity it confers to the user. These two configs do correspond to separate concepts in the code, so it makes sense to have separate classes. We can add a test case to ensure these classes have the same keys so we don't accidentally add a key to one class but not the other.

### Solution 2: Create SAEGroup from a list of SAEConfigs, with helpers for generating permutations
We could have SAEGroup take in a list of SAEConfig objects directly, which would solve the typing issue  as there's no need then for a separate config for the SAEGroup, and would also give the user more direct control over what specific combinations of params they want to be running if they don't want every permutation. We could create some helper functions which can generate this list of SAEConfigs using combinations of values to enable the current behavior too. Something like:

```python
# Create based on combination of params on a base config, to match current behavior
base_config = LanguageModelSAEGroupRunnerConfig(...)
sae_group = SAEGroup.from_combinations(base_config, {
    'l1_coefficient': [1e-3, 1e-4, 1e-5],
    'd_sae': [1000, 5000],
})

# Create directly from a list of SAEConfigs
sae_configs = [
    LanguageModelSAEGroupRunnerConfig(...),
    LanguageModelSAEGroupRunnerConfig(...),
    LanguageModelSAEGroupRunnerConfig(...),
)
sae_group = SAEGroup(sae_configs)
```

I like solution 2 best, as it feels like it more directly addresses what we want a SAEGroup to be, which is simply a collection of SAEs with different params. It allows more flexibility than is possible currently since you can choose combinations of configs that don't need to include every possible permutation of options if you just have a few combinations you want to try out, while still allowing trying out all combinations of params if you'd like instead. It also avoids the DRY issues of having to create a separate, nearly identical config class, and the `SAEGroup.from_combinations()` method makes it really obvious what param combinations are being tried out since it lists them specifically, instead of needing to skim the whole config looking for items in lists.

chanind:
This PR adds optional list typing for config params that can take a list of values for SAEGroup construction, and enables proper config typing for `ActivationsStore` and `SAEGroup` classes.

In #44, I originally tried to remove the config from `SAEGroup` entirely to avoid having the same config class passed to both SAE and SAEGroup, since SAE can't have any list values, but SAEGroup can. I think I was overthinking this, and just explicitly setting the SAEGroup fields that can be lists as optional lists, and adding some assertions in SAE construction that these cfg values are not lists seems to solve the problem. I also set some of these types on the SAE class itself so we can read the values from the SAE object instead of from `sae.cfg` to avoid needing to keep asserting that things aren't lists.

Enabling typing for the `ActivationsStore` cfg revealed that the cache runner config was missing some required params that ActivationsStore needs, so this PR moves those shared params into the base `RunnerConfig` class.

fixes #44

### DIFF
--- a/sae_training/activations_store.py
+++ b/sae_training/activations_store.py
@@ -1,5 +1,7 @@
+from __future__ import annotations
+
 import os
-from typing import Any, Iterator, cast
+from typing import Any, Iterator, Literal, TypeVar, cast
 
 import torch
 from datasets import (
@@ -12,6 +14,11 @@ from datasets import (
 from torch.utils.data import DataLoader
 from transformer_lens import HookedTransformer
 
+from sae_training.config import (
+    CacheActivationsRunnerConfig,
+    LanguageModelSAERunnerConfig,
+)
+
 HfDataset = DatasetDict | Dataset | IterableDatasetDict | IterableDataset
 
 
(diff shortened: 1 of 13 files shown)


jbloomAus:
Thanks! Seems good.

chanind:
I agree the base runner config is confusing, since it seems like it's an implicit ActivationsCache config.

If I understand correctly, we'd remove the base `RunnerConfig` class, and just have a `LanguageModelSAERunnerConfig` and `CacheActivationsRunnerConfig`. Then, for `ActivationsCache`, we could just have the constructor take in the params it needs explicitly, which will be passed in whenever an `ActivationsCache` is constructed. This would be more verbose, but also a lot more clear / obvious what's going on. Is this what you're thinking?

chanind:
Or alternatively, maybe we could create an explicit `ActivationsCacheConfig` class, and have both `LanguageModelSAERunnerConfig` and `CacheActivationsRunnerConfig` extend it. Or we could have `ActivationsCacheConfig` just be a protocol which `LanguageModelSAERunnerConfig` and `CacheActivationsRunnerConfig` both implement.

chanind:
I updated the PR to split apart the runner configs and remove the base class. I initially tried using a Protocol to allow ActivationsStore to directly take in either config, but it felt awkward and I feel like it might lead to confusing typing errors if users aren't familiar with Protocols in Python. That commit is here for reference: https://github.com/chanind/mats_sae_training/commit/d01941c987998b29ffab8bde3263bf894e9a90f2. Instead, the current PR just has the `ActivationsStore` class take in all the params it needs explicitly in the constructor, but adds a `ActivationsStore.from_config()` helper to make it easy to create a store given either runner config if needed.

jbloomAus (commented):
Looks great, thanks!

PR #62, opened 2024-04-01
removing truncation in activations store data loading

chanind:
Currently, when `ActivationsStore` loads the next line of tokens while filling in batches, it truncates the line to the tokenizer max len (1024 tokens for gpt2). This truncation is not necessary, however, since we reshape the tokens that into batches of length `context_size` anyway, regardless of how long the original line of tokens is. As a result, we are likely cutting off a lot of the train dataset text when loading it into the ActivationsStore.

This PR addresses the issue by removing truncation from the tokenization step, so we always tokenize the full length of the training text line.


chanind:
CI is failing for this PR because the `eindex` dependency changed its name to `eindex-callum`. This issue is fixed in https://github.com/jbloomAus/mats_sae_training/pull/63.

jbloomAus:
This truncation was actually intentional to increase the variability in the
tokens being used to train the SAE. We could do an AB test to see if it's
worth it.

chanind:
ah interesting - if it's intentional I'll close this PR. I do feel like the behavior feels unexpected though - maybe we could add a param to the config for something like `max_tokens_per_training_sample` or something like that, so the behavior is more explicit? It seems like it could be decoupled from the tokenizer max_len as well, since there's no real reason why the max len of the tokenizer would necessarily be the correct length to increase token variability.

jbloomAus:
Agreed! We should think about this further. Don't close the pr for now.

chanind:
OK - re-opening this for now

chanind:
OK - re-opening this for now

jbloomAus:
@chanind how did we not merge this?? We should merge it. If you get a sec can you please rebase? Let me know as I'll do it if needed.

chanind:
my bad, updating this PR now and will merge once everything passes

PR #63, opened 2024-04-01
removing unused eindex depencency

chanind:
CI is [currently broken](https://github.com/jbloomAus/mats_sae_training/actions/runs/8509749352/job/23305929241?pr=62) because the `eindex` dependency just changed its name to `eindex-callum` (see [commit](https://github.com/callummcdougall/eindex/commit/695dbb9e9a2a3224c28485fb6e0b8b92c274dcbd)), so our dependencies are now incorrect and we can't install our project. It looks like this dependency is never used in our project anyway, so this PR just removes it as a dependency entirely.


jbloomAus:
Thanks. We used to use it but stopped.

PR #64, opened 2024-04-03
adding lr_decay_steps and refactoring get_scheduler

### LINKED ISSUE #61: Add a decay option to `get_scheduler` and refactor to use SequentialLR
As a follow-up to #57, we should refactor the `get_scheduler()` function to make it possible to add a linear warm-up and decay to any scheduler. The [SequentialLR](https://pytorch.org/docs/stable/generated/torch.optim.lr_scheduler.SequentialLR.html#sequentiallr) scheduler should be able to help with this, and should make the function simpler as well.

chanind:
This PR adds a `lr_decay_steps` config option which will linearly decay learning rate to 0 at the end of training. This PR simplifies the `get_scheduler` function by removing `constantwithwarmup`, `linearwarmupdecay`, and `cosineannealingwarmup` options. Instead, any scheduler can now have a linear warmup and linear decay by setting the `lr_warm_up_steps` and `lr_decay_steps`.

This also required changing some of the options passed to `get_scheduler`. This function now takes in `lr` as well, since there's a discrepancy with how torch's cosine annealing scheduler works compared to how the other schedulers. Specifically, it seems that torch's cosine annealing actually sets the final LR to the value specified by eta_min, but the other schedulers don't know what the current learning rate is and instead can only set the LR to a portion of its original value. In order to address this discrepancy, we need to know the LR so we can calculate what the expected end LR should be as a portion of the base LR. This also has a nice side-effect of meaning that the `lr_end` value is the actual final LR, instead of a portion of the original LR which felt unexpected.

closes #61

### DIFF
--- a/sae_training/config.py
+++ b/sae_training/config.py
@@ -64,9 +64,10 @@ class LanguageModelSAERunnerConfig(RunnerConfig):
     lp_norm: float = 1
     lr: float = 3e-4
     lr_scheduler_name: str = (
-        "constantwithwarmup"  # constant, constantwithwarmup, linearwarmupdecay, cosineannealing, cosineannealingwarmup
+        "constant"  # constant, cosineannealing, cosineannealingwarmrestarts
     )
     lr_warm_up_steps: int = 500
+    lr_decay_steps: int = 0
     train_batch_size: int = 4096
 
     # Resampling protocol args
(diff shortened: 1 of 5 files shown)


jbloomAus:
Looks great! thanks!

PR #65, opened 2024-04-03
passing accidentally overlooked scheduler opts

chanind:
This PR is a followup from #61. I realized after merging that I forgot to pass the `lr_decay_steps` option from the config through to the actual `get_scheduler()` function 🤦‍♂️. This PR fixes that, and also removes all the default values from the `get_scheduler()` function so the type checker will error if we forget to pass through a value from the config for any param to `get_scheduler()` moving forward.

### DIFF
--- a/sae_training/train_sae_on_language_model.py
+++ b/sae_training/train_sae_on_language_model.py
@@ -216,13 +216,16 @@ def _build_train_context(
     n_frac_active_tokens = 0
 
     optimizer = Adam(sae.parameters(), lr=sae.cfg.lr)
+    assert sae.cfg.lr_end is not None  # this is set in config post-init
     scheduler = get_scheduler(
         sae.cfg.lr_scheduler_name,
         lr=sae.cfg.lr,
         optimizer=optimizer,
         warm_up_steps=sae.cfg.lr_warm_up_steps,
+        decay_steps=sae.cfg.lr_decay_steps,
         training_steps=total_training_steps,
-        lr_end=sae.cfg.lr / 10,  # heuristic for now.
+        lr_end=sae.cfg.lr_end,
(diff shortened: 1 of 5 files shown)


sae_training/train_sae_on_language_model.py
```diff
@@ -216,13 +216,16 @@ def _build_train_context(
...
     optimizer = Adam(sae.parameters(), lr=sae.cfg.lr)
+    assert sae.cfg.lr_end is not None  # this is set in config post-init
     scheduler = get_scheduler(
         sae.cfg.lr_scheduler_name,
         lr=sae.cfg.lr,
         optimizer=optimizer,
         warm_up_steps=sae.cfg.lr_warm_up_steps,
+        decay_steps=sae.cfg.lr_decay_steps,
         training_steps=total_training_steps,
-        lr_end=sae.cfg.lr / 10,  # heuristic for now.
```
chanind, line 225: I also moved this `lr_end` into an explicit config option, setting the default value to `lr / 10`. I figured it's more clear to the end-use what's going on that way.

sae_training/train_sae_on_language_model.py
```diff
@@ -216,13 +216,16 @@ def _build_train_context(
...
+        lr_end=sae.cfg.lr_end,
+        num_cycles=sae.cfg.n_restart_cycles,
```
chanind, line 228: it looks like we didn't have a way to configure the `num_cycles` param for the `cosineannealingwarmrestarts`, so I added an option for that to the config as well. Otherwise `cosineannealingwarmrestarts` just does the same thing as `cosineannealing`.

jbloomAus: all good.

PR #66, opened 2024-04-05
feat: setting up sae_lens package and auto-deploy with semantic-release

### LINKED ISSUE #10: Publish on PyPI
Could this library be published on PyPI so it could be installed via `pip install` instead of needing to clone the git repo? It would make it much easier to install and use this library, as publishing on PyPI is usually the preferred way of sharing Python code.

The quickest path to this would be to do the following:
1. Move the library code into a folder with the same name as the library. If you want to call the library `mats_sae_training` so you can `pip install mats_sae_training`, then create a `mats_sae_training` folder and the `sae_training` and `sae_analysis` folders should be moved into that folder. Imports will then be `from mats_sae_training.sae_training import ...`.
2. Use a modern packing tool ([Poetry](https://python-poetry.org/), [PDM](https://pdm-project.org/latest/), or [Hatch](https://hatch.pypa.io/latest/) are all good choices)
3. Remove the `requirements.txt` and instead split dependencies between the main dependencies and dev dependencies in `pyproject.toml`.
4. Remove `.gitmodules` - it seems this isn't used anyway.

From there, you just need to run the `publish` command for the corresponding packing library you choose from above (e.g. `poetry publish` for poetry, or `pdm publish` for pdm, `hatch publish` for hatch), and you're basically done.

You can also set up automatic deployment with Github Actions using [Semantic release](https://semantic-release.gitbook.io/semantic-release/) so deployment happens automatically when committing to `main`, but that's not strictly required.

I can set up a PR with these changes if it's helpful.

chanind:
This PR renames the package to `sae-lens` to ready it for publication on PyPI. All imports are now from `sae_lens`. The `sae_training` and `sae_analysis` packages are now accessed at `sae_lens.training` and `sae_lens.analysis`, respectively. The most commonly imported items from the `training` package are also exported at the top-level, so it's possible to run `from sae_lens import SparseAutoencoder`, for example.

In addition, this PR sets up the library to auto-deploy on commit to main based on [semantic-release](https://semantic-release.gitbook.io/semantic-release). This means that the commit message to main will determine how the version is bumped in a deploy, or whether a deploy happens at all. For example, the following commit messages will bump the version number as follows:

- `chore: updating docs`  no version bump
- `fix: the main class no longer errors on osx` patch version bump (1.2.3 -> 1.2.4)
- `feat: added a new function` minor version bump (1.2.3 -> 1.3.0)
- `feat: redid everything BREAKING CHANGE: everything is different` major version bump (1.2.3 -> 2.0.0)

I tried to update all the docs and jupyter notebooks to reflect the new import pattern, but it's possible I missed something somewhere.

This PR adds a custom unpickler method so that currently existing pickled SAEs on huggingface can still be loaded with `torch.load()` in the session loader.

closes #10


jbloomAus:
Amazing, thanks so much!

PR #68, opened 2024-04-06
fix: hotfix updating saevis version to use pypi

chanind:
This PR uses the `sae-vis` version on PyPI so that we can hopefully successfully publish to PyPI.

This has no typing errors, but I commented out a few places in the neuronpedia runner, and didn't manually try creating a dashboard so it's possible it's not working well in practice.

### DIFF
--- a/sae_lens/analysis/neuronpedia_runner.py
+++ b/sae_lens/analysis/neuronpedia_runner.py
@@ -9,8 +9,15 @@ import time
 import numpy as np
 import torch
 from matplotlib import colors
+from sae_vis.data_config_classes import (
+    ActsHistogramConfig,
+    Column,
+    FeatureTablesConfig,
+    SaeVisConfig,
+    SaeVisLayoutConfig,
+    SequencesConfig,
+)
 from sae_vis.data_fetching_fns import get_feature_data
-from sae_vis.data_storing_fns import FeatureVisParams
 from tqdm import tqdm
 
 from sae_lens.training.session_loader import LMSparseAutoencoderSessionloader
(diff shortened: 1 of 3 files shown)


sae_lens/analysis/neuronpedia_runner.py
```diff
@@ -269,49 +294,52 @@ def run(self):
...
-                            feature.left_tables_data.correlated_neurons_indices
-                        )
-                        feature_output["correlated_neurons_l1"] = self.round_list(
-                            feature.left_tables_data.correlated_neurons_l1
+                            feature.feature_tables_data.correlated_neurons_indices
                         )
+                        # TODO: this value doesn't exist in the new output type, commenting out for now
+                        # there is a cossim value though - is that what's needed?
+                        # feature_output["correlated_neurons_l1"] = self.round_list(
+                        #     feature.feature_tables_data.correlated_neurons_l1
```
chanind, line 313: I just commented this out since I didn't know what the new field is called

sae_lens/analysis/neuronpedia_runner.py
```diff
@@ -269,49 +294,52 @@ def run(self):
...
+                        vocab_dict, feature.logits_table_data.top_token_ids
                     )
                     feature_output["pos_values"] = top10_logits
 
-                    feature_output["frac_nonzero"] = (
-                        feature.middle_plots_data.frac_nonzero
-                    )
+                    # TODO: don't know what this should be in the new version
+                    # feature_output["frac_nonzero"] = (
+                    #     feature.middle_plots_data.frac_nonzero
```
chanind, line 339: here as well, not sure what the new name for this is, if it exists in the new version

jbloomAus:
Thanks!

PR #71, opened 2024-04-07
Addressing notebook issues 

weissercn:
Addressing two issues with logits_lens_with_features.ipynb

1. Backwardscompatible unpickler was not used leading to failure to load
2. Missing nltk import and loading of averaged_perceptron_tagger
3. Missing key for df_enrichment_score 
4. Fixed a graph name


jbloomAus:
Thanks! Running CI and then if it passes I'll accept.

weissercn:
There was an issue with linting, which I have addressed.

jbloomAus:
thanks :)

PR #75, opened 2024-04-08
Added support for Fengyuan Hu's implementation of sparse autoencoder.

HuFY-dev:
Added options in config to use `sae_type="normalized_sae"` and set `noise_scale` for [this implementation of SAE](https://github.com/HuFY-dev/sparse_autoencoder/tree/dev). Default is set to `sae_type="sae"` and `noise_scale=0.0` which should behave exactly the same as the old version.

### DIFF
--- a/sae_lens/training/sparse_autoencoder.py
+++ b/sae_lens/training/sparse_autoencoder.py
@@ -35,6 +35,8 @@ class SparseAutoencoder(HookedRootModule):
     hook_point_layer: int
     dtype: torch.dtype
     device: str | torch.device
+    noise_scale: float
+    sae_type: str
 
     def __init__(
         self,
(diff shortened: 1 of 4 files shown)


sae_lens/training/sparse_autoencoder.py
```diff
@@ -180,26 +196,28 @@ def initialize_b_dec_with_mean(self, all_activations: torch.Tensor):
 
     @torch.no_grad()
     def set_decoder_norm_to_unit_norm(self):
-        self.W_dec.data /= torch.norm(self.W_dec.data, dim=1, keepdim=True)
```
jbloomAus, line 183: I think the if statement should go outside of this function (would be much more intuitive)

sae_lens/training/sparse_autoencoder.py
```diff
@@ -180,26 +196,28 @@ def initialize_b_dec_with_mean(self, all_activations: torch.Tensor):
...
+            self.W_dec.data /= torch.norm(self.W_dec.data, dim=1, keepdim=True)
 
     @torch.no_grad()
     def remove_gradient_parallel_to_decoder_directions(self):
         """
         Update grads so that they remove the parallel component
             (d_sae, d_in) shape
         """
-        assert self.W_dec.grad is not None  # keep pyright happy
+        if self.sae_type == "sae":
```
jbloomAus, line 208: again, if statement should be outside of this

sae_lens/training/sparse_autoencoder.py
```diff
@@ -114,7 +124,13 @@ def forward(self, x: torch.Tensor, dead_neuron_mask: torch.Tensor | None = None)
             )
             + self.b_enc
         )
-        feature_acts = self.hook_hidden_post(torch.nn.functional.relu(hidden_pre))
+        if self.sae_type == "sae":
```
jbloomAus, line 127: I wonder if there's a cleaner way to do the forward pass that doesn't involve if statements (i expect this to slow the code down if we leave the if statements in)

jbloomAus:
- I think installing your code as a dependency may cause issues for pip install (not sure if we currently test this in CI).
- I think normalized SAE sounds too generic. Maybe making an arg for the activation function and calling it tanh_relu would be clearer?

HuFY-dev:
> * I think installing your code as a dependency may cause issues for pip install (not sure if we currently test this in CI).
> * I think normalized SAE sounds too generic. Maybe making an arg for the activation function and calling it tanh_relu would be clearer?

I tried to do the loading without installing my code, but the saved file is not a simple state dict and can't be loaded directly. I'll have to see what I can do with it.

PR #76, opened 2024-04-09
perf: improving CI speed

chanind:
This PR improves the speed of CI to take around ~2 min when all caches hit. This is accomplished by caching as many dependencies as possible, but most of the speedup is from speeding up tests.

The dependencies cached are the following:
- poetry
- python dependencies
- huggingface models

Tests were slow mainly because `ActivationsStore` fills up its store on initialization, and the default sizes of the this store in the test config were pretty large. So, every time an `ActivationsStore` was created we were running a lot of inference on models on CPU in CI.

This PR also moves all the `ActivationsStore` tests on real models/datasets into 1 test, since loading real models and data is a big fixed cost. The new test has all the same assertions as the previous 3 tests, but runs a lot faster than reloading the `ActivationsStore` and data 3 times for 4 different models/datasets.

This PR also refactors the code that tests out loading the pre-trained GPT2 SAE from huggingface. Previously, this was done by running `LMSparseAutoencoderSessionloader.load_session_from_pretrained()`, since the logic for handling old dict-based loading was handled there. However, this function also initializes an `ActivationsStore`, which then tries to immediately fill up a batch of activations which is really slow.

Instead, the functionality to handle loading a `SAEGroup` from pretrained is now all contained in `SAEGroup.from_pretrained()`. This means that calling `SAEGroup.from_pretrained()` will now always return a `SAEGroup` instance or will error, while previously it could return a dict or really anything that was contained in the pickled instance. Moving this logic into `SAEGroup.from_pretrained()` means we can test just loading the pre-trained SAE from huggingface without also needing to inialize an `ActivationsStore` and load a big batch of data. Also I feel like it's a more natural place for to handle coercing old loaded files into a `SAEGroup`, since IMO it felt unexpected that calling `SAEGroup.from_pretrained()` could return things that aren't SAEGroups.

**NOTE** this PR also removes the initial filling of the buffer in `ActivationsStore.__init__`, instead making this initial filling always happen lazily when a buffer is requested instead. I don't see a case where this wouldn't be preferable to loading up-front, especially for users who just want to use a SAE not do training, and should speed up tests as well.


jbloomAus:
Apologies for any work that is created by the PR I'm making now (https://github.com/jbloomAus/SAELens/pull/78).

Some thoughts:
- loading / saving from disk  has changed decently so it will be important to check those changes are all compatible. 
- I think we should make it possible to load the activations store but not immediately fill the buffer. 
- The same way you can override part of the config (like the device) when loading from pretrained, we should probably let users override the number the store batch size and  the n_batches_in_buffer.

chanind:
Yeah I agree that loading a batch of activations on init isn't ideal. I feel like we should do that lazily when the activations are first requested instead to avoid unnecessary work. Also agree that these params should be changeable when loading the store - there's no reason why someone fine-tuning a SAE further needs to use the same batch sizes and stuff as the original training.

chanind:
I just made the initial batch loading lazy, so the `ActivationsStore` won't actually do any processing until a batch is actually requested. Otherwise, this should be good to go

jbloomAus:
Love your work! Great stuff!

PR #77, opened 2024-04-10
Added `tanh_sae` support & adjusted implementation based on comments.

HuFY-dev:
Following suggestions from the previous PR, I change the implementation slightly (and removed the if statement from the forward pass). However I still can't load the model without using my library, so pip-installing from GitHub is still needed. The code passed the style and unit tests on PR to my main (should be the same workflow).


jbloomAus:
Thanks! Will review properly shortly! Sorry for the delay.

jbloomAus:
Apologies for the delay. I think it would be good to avoid installing your library from github and might be able to invest some time in supporting this soon. Neuronpedia has some other SAE's which we need to upload / support that unfortunately take priority (but we'll try to get to this ASAP!)

chanind:
Adding an option for activation function, and for adding noise both make sense as options. It shouldn't be necessary to load the `sparse_autoencoder` package, though, or change the Python version requirements in Pyproject.toml.

Could you save your SAE and upload to huggingface using safetensors instead of `torch.save()`? That would let us just load the weights dict without needing to have access to the original class where the weights come from.

I'm going to modify this PR to rename some of the options and the pyproject.toml to get this merged. Apologies for the delay with this!

chanind:
I changed the config options as follows:
- `activation_fn`: either "relu" (default) or "tanh-relu". We can add more options in the future
- `normalize_sae_decoder`: bool, (`True` by default).
- removed `sae_type` in favor of the above options.
- `noise_scale` is still the same.

I don't think the notebook will work at present, but I'd be up for removing the notebook from this PR to merge these new options, and then the notebook can be added back in a PR later if it's possible to load the model state-dict from huggingface.

jbloomAus:
Awesome, let's hold off for a day in case @HuFY-dev  can make the HF repo.

HuFY-dev:
I uploaded to the model statedict and config to `HuFY-dev/tanh_sae` on HuggingFace. You can use either the `.pt` and the `.safetensors` model and both should give you state dicts now. The config in a separate file with the same format as earlier implementations.

PR #79, opened 2024-04-11
Mamba support

Phylliida:
Changes:

cfg has the following new fields:
- model_class: By default this is HookedTransformer, but can be changed to HookedMamba
- hook_point_eval: Before this was hardcoded
- model_kwargs: passed into all calls to the model (useful because you can enable the kernel support with `{"fast_ssm": True, "fast_conv": True}`

Fixes the get_name() function to still provide valid file paths when model paths have / in them (the typical model used is state-spaces/mamba-370m)

Prevent divide by zero in evals


jbloomAus:
Thanks @Phylliida  for making mamba-lens a package https://pypi.org/project/mamba-lens/ 
@chanind may have time to make the required changes :) Otherwise I'll get to it soon.

chanind:
Ah excellent! I'll work on this today or tomorrow

chanind:
@Phylliida I added a test case for loading a mamba model, but this test is currently failing. I think this is on `mamba-lens`'s side. This test is just calling `HookedMamba.from_pretrained("state-spaces/mamba-370m", device="cpu")`.

The error is `ValueError: mutable default <class 'mamba_lens.HookedMamba.MambaInitCfg'> for field initializer_cfg is not allowed: use default_factory`

Phylliida:
@chanind ok that should be fixed now, I released a new version of mamba-lens on pypi

chanind:
@Phylliida It looks like it's still not fixed. The issue is the following line: https://github.com/Phylliida/MambaLens/blob/main/mamba_lens/HookedMamba.py#L294C37-L294C49

Specifically, this has the following:
```python
initializer_cfg: MambaInitCfg = MambaInitCfg()
```

This should be changed to
```python
from dataclasses import field

initializer_cfg: MambaInitCfg = field(default_factory=MambaInitCfg)
```

It looks like this only errors in Python 3.11+, but it's good practice in general to never set mutable defaults on dataclasses without using `default_factory`. The problem is that the way this is set, if you create 2 instances of `MambaCfg`, they will share the same `initializer_cfg` object, so changing one `initializer_cfg` will change both of their `initializer_cfg`s.

e.g.
```python
cfg1 = MambaCfg()
cfg2 = MambaCfg()

cfg1.initializer_cfg.dt_scale = 7.0
print(cfg2.initializer_cfg.dt_scale) # this will print 7.0 as well now
```

Phylliida:
very sorry I thought I had made that change but for some reason it didn't go through, try it now

chanind:
Looks like that worked! Just merged

PR #80, opened 2024-04-12
bugfix - minimum viable updates to tutorial notebook

wllgrnt:
While running through the tutorial notebook I encountered a couple of runtime errors - these are the smallest set of updates that get you something runnable, and generate html that looks like:

image

I'm sure given the dev speed this'll be out of date again soon, but this hopefully this suffices in the interim.


jbloomAus:
@wllgrnt thank you! Sorry for unreliability. Aiming to achieve more stability moving forward.

PR #81, opened 2024-04-15
feat: add class to patch HookedTransformer with SAE

dtch1997:
Adds an `SAEPatcher` class: 
- `get_forward_hook()` and `get_backward_hook()` simplify patching a model computation graph with SAE features and error terms. Example usage given in tests. 
- `get_node_values()` and `get_node_grads()` can be used to retrieve the activations and gradients subsequently. 
- Test that patching the model in this way does not alter the forward pass (loss, activations) or the backward pass (gradients). 

Other changes
- Adds `encode` and `decode` methods for `SparseAutoencoder` based on existing implementation


sae_lens/training/sparse_autoencoder.py
```diff
@@ -102,7 +102,10 @@ def __init__(
 
         self.setup()  # Required for `HookedRootModule`s
 
-    def forward(self, x: torch.Tensor, dead_neuron_mask: torch.Tensor | None = None):
+    def encode(
+        self, x: torch.Tensor, return_hidden_pre: bool = False
```
chanind, line 106: nit: I feel like the style of having different return types based on an input param should be avoided, as it breaks typing and makes it hard to know what methods are returning. Could we just return 1 thing from this method? It seems unnecessary to return the hidden_pre anyway IMO since this is already hooked, so it's easy to access that value as needed using the hook

sae_lens/toolkit/sae_patcher.py
```diff
@@ -0,0 +1,135 @@
...
+        """Forward hook to patch the SAE into the computational graph
+
+        orig: Original activations
+        """
+
+        a_orig = orig
+        z_sae = self.sae.encode(a_orig)
+        # keep pyright happy
+        assert isinstance(z_sae, torch.Tensor)
+        a_sae = self.sae.decode(z_sae)
```
chanind, line 62: I would worry about using a different forward pass calculation when patching, since it seems like it could result in subtle bugs if we change the implementation of `forward()` in the model. Can't we use the normal `.forward()` here, as it already returns the hidden activations of the SAE?

sae_lens/toolkit/sae_patcher.py
```diff
@@ -0,0 +1,135 @@
...
+
+    @property
+    def sae_nodes(self) -> Float[torch.Tensor, "n_batch n_token (d_sae + d_model)"]:
+        return torch.cat([self.sae_feature_acts, self.sae_errors], dim=-1)
+
+    def _backward_hook_fn(
+        self,
+        orig: Float[torch.Tensor, "n_batch n_token d_model"],
+        hook: HookPoint,
+    ) -> tuple[Float[torch.Tensor, "n_batch n_token d_model"]]:
```
chanind, line 86: Why is this function needed? It looks like it's just the identity

sae_lens/toolkit/sae_patcher.py
```diff
@@ -0,0 +1,135 @@
...
+        z_sae = self.sae.encode(a_orig)
+        # keep pyright happy
+        assert isinstance(z_sae, torch.Tensor)
+        a_sae = self.sae.decode(z_sae)
+        a_err = a_orig - a_sae.detach()
+
+        # Track the gradients
+        assert z_sae.requires_grad
+        z_sae.retain_grad()
+        assert a_err.requires_grad
```
chanind, line 68: nit: it shouldn't be necessary to require gradients in order to apply a SAE to the model. This is only necessary for attribution patching. Maybe this should be an input option?

sae_lens/toolkit/sae_patcher.py
dtch1997: The idea here is that, when we patch the model with an SAE at an activation, there are actually two copies of the activation we care about: $a^u$, the copy which is fed into the SAE, and $a^d$, the copy which comes out of it. 

The values of $a^u$ and $a^d$ are the same, but the specific implementation means that gradients from $a^d$ do not get passed to $a^u$. (Concretely, we called `detach()` somewhere in there).

To fix this, we need a backward hook which directly passes the gradient of $a^d$ to $a^u$. This hooks does that, since the input is the gradient of $a^d$ and the output is the new gradient of $a^u$.

dtch1997: See Marks' paper, appendix A3, 'Stop gradients` and `Passthrough gradients` for more explanation

dtch1997: Yeah, basically it boils down to the fact that if we don't have this then everything before that layer won't receive gradient!

sae_lens/training/sparse_autoencoder.py
dtch1997: I was mainly motivated by wanting to be able to re-use this functionality in `forward`. happy to modify this if you think there's a better way

sae_lens/toolkit/sae_patcher.py
dtch1997: yeah, I could add a flag in the constructor to support inference w/o gradients

jbloomAus:
@dtch1997 per other merges. 3.0.0 now supports this functionality.

PR #86, opened 2024-04-16
chore: re-enabling isort in CI

chanind:
This PR re-enables isort in CI. If this is still an issue in the future, we should just remove isort entirely.


chanind:
https://stackoverflow.com/questions/73471822/isort-sorting-things-differently-when-running-locally-than-in-ci-other-machines this question makes me thing the inconsistencies with isort might have stemmed from the `wandb` folder locally making isort think `wandb` is a local module, and sorting it incorrectly. I added `src_paths` to the isort config, which should hopefully fix this issue

PR #87, opened 2024-04-17
Adds function that converts old .pt pretrained SAEs to new folder format

evanhanders:
Title basically says it all:

- Adds `convert_old_to_modern_saelens_format` to toolkit/pretrained_saes.py
- Adds test of above function
- Fixes a bug for non-mps loads in legacy loader


jbloomAus:
Apologies, to pass CI you need to run the linting / formatting checks (explained here: https://jbloomaus.github.io/SAELens/about/contributing/)

Have made a note to make this more prominent in github.

PR #90, opened 2024-04-18
WIP: Better evals

derpyplops:
Statistics
[] Dead feature count

Qualitiative
[] Does the model lose specific capabilities (eg: IOI) when we use the SAE?

Viz
[] Generate evals table
[] Generate Feature Density Histograms

Reproduceability
[] Notebook can run off the bat in colab

Extensions (probably future PR)
[] How does performance vary for SAEs in a set (eg: different layers or different expansion factors or L0s)
[] How do the logit lens statistics shift between features?


jbloomAus:
I think this is unlikely to be merged but I do want to point people towards it in the future. There were some things that needed fixing but there were some useful improvements.

derpyplops:
What things needed fixing? I could work on those if it's wanted

PR #95, opened 2024-04-20
Make load_state_dict use strict=False

neelnanda-io:
Your library is too strict, which creates errors if loading state dicts that don't eg contain a scaling factor. I added a quick fix that won't raise errors when loading a state dict that only contains some parameters. This runs the risk that someone may accidentally load a state dict with too few parameters, which is bad? But seems pretty unlikely to me

### DIFF
--- a/sae_lens/training/sparse_autoencoder.py
+++ b/sae_lens/training/sparse_autoencoder.py
@@ -312,7 +312,13 @@ class SparseAutoencoder(HookedRootModule):
 
         # Create an instance of the class using the loaded configuration
         instance = cls(cfg=state_dict["cfg"])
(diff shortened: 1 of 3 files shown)


hijohnnylin:
Perhaps checking that the state dict’s keys is a superset of a set of required keys?

I recently made a very similar change to sae_vis here: https://github.com/callummcdougall/sae_vis/pull/39/files

jbloomAus:
Hmm. So this is resulting from the decoder fine-tuning changes which added a new state_dict parameter to SAEs. This parameter doesn't effect the output if unused. 

I think we should:
- [ ] explicitly check whether the extra weight is the scale parameter. 
- [ ] deliberately allow strict = false if that's the only difference.

PR #106, opened 2024-04-25
feat: Store state to allow resuming a run

Phylliida:
# Description

- Stores n_training_steps n_training_tokens, started_fine_tuning, checkpoint_paths and rng state in SAETrainingRunState
- Writes ctx to file (this includes optimizer and scheduler state)
- Code for loading ctx from file (correctly attached to sae)
- Saves ActivationStore (in particular, _storage_buffer and a new variable n_dataset_processed)
- ActivationStore has n_dataset_processed that keeps track of how far along in dataset
- When ActivationStore is loaded, it is fastforwarded
- config has a resume parameter that looks for the most recent checkpoint and loads it (see lm_runner)
- config has log_activations_store_to_wandb boolean
- config has log_optimizer_state_to_wandb boolean
- for these booleans, if false, activations store and optimizer state will still be saved, just locally, it won't be uploaded to wandb (these are quite big, so don't want to spam wandb too much)
- final folder is now final_{num steps} (needed for finding most recent)
- added test in tests/unit/training/test_train_sae_on_language_model.py for saving and loading
- added wandb_id param to config (optional, used to resume the same wandb run)

# Summary

I usually do something like this:

```python
if not 'JOB_NAME' in os.environ or os.environ['JOB_NAME'] is None or os.environ['JOB_NAME'] == "":
    raise ValueError(f"Need to set environment variable JOB_NAME")

job_name_text = os.environ['JOB_NAME'] + ".txt"

if os.path.exists(job_name_text):
    with open(job_name_text, "r") as f:
        wandb_id = f.read().strip()
    print(f"resuming from {wandb_id}")
    resume = True
else:
    wandb_id = wandb.util.generate_id()
    with open(job_name_text, "w") as f:
        f.write(wandb_id)
    print(f"starting new run {wandb_id}")
    resume = False
```
and then I set `cfg.wandb_id = wandb_id`
and `cfg.resume = resume`

this will generate an id on your first run, and if interrupted (ctrl-c, sigint, sigterm) an additional checkpoint will be saved

then when you run again with the same id, the most recent checkpoint will then automatically be used and the wandb run can be resumed

## Type of change

- [x ] New feature (non-breaking change which adds functionality)


chanind (commented):
This PR looks reasonable to me! It's hard to say with certainty that the resuming logic doesn't have suble bugs without a test that asserts that the resulting trained SAE is identical when train with and without resuming, or without more intricate unit-tests on the indivual moving parts of the PR, but in practice that might be hard to do. It's also our fault for not having tests on the LM runner yet. It seems like this PR shouldn't break any existing functionality as far as I can tell, so I'd support merging this if @jbloomAus is OK with it as well.

Phylliida:
Yes I also wasn't sure how much non-determinism training had, and even if none that'd require some way of interrupting runs via code (seperate thread sent a sigint would work but would be gross and timing based)

PR #108, opened 2024-04-27
API for generating autointerp + scoring for neuronpedia

hijohnnylin:
# Description

This adds the ability to pull a neuronpedia feature, do autointerp and scoring on it locally (with openai key that's stored locally), then uploads it to neuronpedia.

## Type of change

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -8,7 +8,6 @@ packages = [{include = "sae_lens"}]
 
 [tool.poetry.dependencies]
 python = "^3.10"
-neuron-explainer = { git = "https://github.com/hijohnnylin/automated-interpretability.git", subdirectory = "neuron-explainer" }
 transformer-lens = "^1.14.0"
 transformers = "^4.38.1"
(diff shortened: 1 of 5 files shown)


pyproject.toml
```diff
@@ -8,6 +8,7 @@ packages = [{include = "sae_lens"}]
 
 [tool.poetry.dependencies]
 python = "^3.10"
+neuron-explainer = { git = "https://github.com/hijohnnylin/automated-interpretability.git", subdirectory = "neuron-explainer" }
```
chanind, line 11: I don't think we can have dependencies on a git repo while still being publishable on PyPI. This would also need to be published on PyPI as well.

hijohnnylin: ah thanks! converting this PR to draft until i do this

hijohnnylin: alrighty it's done and ready for review

chanind (approved):
👍 lgtm!

PR #115, opened 2024-04-30
fix: fixing cache_activations_runner and adding a test

chanind:
# Description

This PR fixes an issue in `cache_activations_runner` where the runner will always error since it requires a `cached_activations_path` to be set, but the `ActivationsStore` will error if an empty `cached_activations_path` is passed to it. This PR adds a basic test case to the `cache_activations_runner()` function as well.

This PR doesn't change any other implementation of `cache_activations_runner()`, but a followup PR will investigate saving a huggingface datasest consisting of activations to disk instead of using `torch.save()`.

## Type of change

# Checklist:

### DIFF
--- a/sae_lens/training/activations_store.py
+++ b/sae_lens/training/activations_store.py
@@ -50,6 +50,9 @@ class ActivationsStore:
             and not cfg.use_cached_activations
         ):
             cached_activations_path = None
+        # don't try to load from the cache for the cache runner, since we're building a new cache
+        if isinstance(cfg, CacheActivationsRunnerConfig):
+            cached_activations_path = None
         return cls(
             model=model,
             dataset=dataset or cfg.dataset_path,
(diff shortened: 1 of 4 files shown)


jbloomAus:
Thanks David. 

- Reminder that we wanted to move to using safetensors or something else that is faster / more efficient for activations cached to disk. I could probably make use of this ASAP, so let me know if/when you think you might get to this!

chanind:
Apologies, I didn't mean to imply that this PR means not improving cache loading, I just figured it was worth fixing this since it's currently broken. I'll close this PR for now since it sounds like it's not worth fixing this without also improving performance.

jbloomAus:
Oh, I meant to accept this! I understood it was incremental. No need to close.

chanind:
Ah sorry for the misunderstanding, just re-opened!

PR #119, opened 2024-05-02
Fix: sparsity norm calculated at incorrect dimension.

HuFY-dev:
# Description

For L1 this does not effect anything as essentially it's calculating the abs() and average everything. For L2 this is problematic as L2 involves sum and sqrt. Unexpected behaviors occur when x is of shape (batch, sen_length, hidden_dim).

## Type of change

# Checklist:

### Performance Check.

- [ ] L0
- [ ] CE Loss
- [ ] MSE Loss

### DIFF
--- a/sae_lens/training/sparse_autoencoder.py
+++ b/sae_lens/training/sparse_autoencoder.py
@@ -198,7 +198,7 @@ class SparseAutoencoder(HookedRootModule):
             )
 
         mse_loss = per_item_mse_loss.mean()
-        sparsity = feature_acts.norm(p=self.lp_norm, dim=1).mean()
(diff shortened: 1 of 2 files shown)


chanind:
This looks like a reasonable change, but would it be possible add a test case which would fail without your fix, but passes with your fix?

HuFY-dev:
I added a test and found that this error only occurs during the eval phase, where the input x is 3D: (batch, seq_length, d_in). However, during eval, l1 loss is not used, so it didn't cause any error to just run the code. In fact, the old code will return a l1 with shape (seq_length) if you give it a 3d input, which is not desirable. This does not affect the training phase, as the input to the training phase is always 2D: (batch, d_in).

PR #120, opened 2024-05-04
Make tutorial notebooks work in Google Colab

nlpet:
# Description

This PR makes a few tutorial notebooks from `tutorials` available in Google Colab by adding dependency installs at the top, so that they can be easily run in Colab. I've tested the notebooks in Google Colab and they all work, `logit_lens_with_features` requires more RAM than is available out of the box on Colab, so I've added a comment near the dependency installs regarding that. I've also tested them locally (on a mac) to ensure they still work and they do. 

Changes included:
- Dependency installs at the top so that code runs on Colab
- Cleared the outputs from running the cells (so that future changes are easier to parse)
- Section in README pointing to tutorial notebooks with a "run on Colab" button
- In `basic_loading_and_analysing`, there is a visualisation involving displaying html files, I've made changes to make that work on Colab too

As far as I am aware this does not affect any core functionality or tests, so I've removed the checklists. I've also excluded some of the notebooks / scripts in `tutorials` as they did not look like full tutorials but I am happy to include any additional ones!


jbloomAus:
@chanind just checking ig this is GTG?

chanind:
Sorry for the delay with this! Awesome work with these notebooks the "open in colab" links are a nice touch as well!

PR #121, opened 2024-05-05
fix: moving non-essential deps to dev

chanind:
# Description

We have a number of dependencies listed as core dependencies in pyproject.toml that look like they should either be in dev, or removed entirely. Everything listed in `[tool.poetry.dependencies]` will be installed by users when they `pip install sae-lens`, so this should only include dependencies that are required to use the library.  For instance, the `mkdocs` stuff is for building the docs website rather than using the library, and there were a number of jupyter-related dependencies that are also not necessary for using the library. For Jupyter, I feel this should not be listed in deps at all, as it should be up to the user to decide if they want to run in Jupyter notebook vs Jupyter lab vs colab vs vs-code Jupyter plugin vs normal python, and our deps shouldn't pick that for them.

## Type of change

# Checklist:


jbloomAus:
@chanind Should be fine to merge once conflicts are resolved.

PR #126, opened 2024-05-07
Add bf16 autocast

tomMcGrath:
# Description

Adds torch autocasting to the SAE train step, improving throughput.

## Type of change

# Checklist:

- [x] I have added tests that prove my fix is effective or that my feature works (see W&B dash)

### Performance Check.

- [x] L0
- [x] CE Loss
- [x] MSE Loss

W&B dashboard showing no performance regression when autocasting to bf16 and moderate speedup/perf-per-time improvement: https://wandb.ai/tmcgrath/autocast%20testing/workspace

More performance could probably be obtained by compiling and autocasting the LLM and maybe by compiling the SAE. Other hyperparameters are also probably better for performance.

### DIFF
--- a/sae_lens/training/config.py
+++ b/sae_lens/training/config.py
@@ -64,6 +64,7 @@ class LanguageModelSAERunnerConfig:
     seed: int = 42
     dtype: str | torch.dtype = "float32"  # type: ignore #
     prepend_bos: bool = True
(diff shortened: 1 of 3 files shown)


tomMcGrath:
BTW I tried adding fp16 support but training performance degraded pretty badly so it seems more like a footgun and it also makes configs hard to serialise.

PR #128, opened 2024-05-08
Change eval batch size

tomMcGrath:
# Description

Surfaces extra options for eval step: number of eval batches and number of sequences per eval batch. Adds an optional parameter to activations_store.get_batch_tokens to make this clean: now get_batch_tokens can get a number of sequences that differs from the value defined at init.

## Type of change

# Checklist:

- [] I have added tests that prove my fix is effective or that my feature works

I've checked these locally but not added tests.

### DIFF
--- a/sae_lens/training/config.py
+++ b/sae_lens/training/config.py
@@ -106,6 +106,10 @@ class LanguageModelSAERunnerConfig:
 
     dead_feature_threshold: float = 1e-8
 
+    # Evals
+    n_eval_batches: int = 10
+    n_eval_seqs: int | None = None  # useful if evals cause OOM
+
     # WANDB
     log_to_wandb: bool = True
     log_activations_store_to_wandb: bool = False
--- a/sae_lens/training/activations_store.py
+++ b/sae_lens/training/activations_store.py
@@ -191,12 +191,12 @@ class ActivationsStore:
             self._dataloader = self.get_data_loader()
         return self._dataloader
 
-    def get_batch_tokens(self):
+    def get_batch_tokens(self, batch_size: int | None = None):
         """
         Streams a batch of tokens from a dataset.
         """
-
-        batch_size = self.store_batch_size
+        if not batch_size:
+            batch_size = self.store_batch_size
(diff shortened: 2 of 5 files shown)


sae_lens/training/config.py
```diff
@@ -106,6 +106,10 @@ class LanguageModelSAERunnerConfig:
 
     dead_feature_threshold: float = 1e-8
 
+    # Evals
+    n_eval_batches: int = 10
+    n_eval_seqs: int | None = None  # useful if evals cause OOM
```
chanind, line 111: nit: is this the batch size used for eval? If so, this would be clearer named as `eval_batch_size` to match the naming of `store_batch_size` and `train_batch_size` params

chanind: ah looks like I must have commented as this was being merged 😅. Not a big issue regardless

jbloomAus: True. Feel free to make a PR for this.

tomMcGrath:
Yeah, the reason I called it this was that the batch size here is the
number of sequences, but the batch size produced for the SAEs is #
sequences * context length. Having an ‘eval_batch_size=4’ in the config
seemed like it would be surprising next to a batch size of (say) 4096.

chanind:
👍 makes sense!

jbloomAus:
@tomMcGrath  @tomMcGrath We previously has "train_batch_size" which is a token batch size for the SAE and a "store batch size" which was a prompts batch size for generating activations. The latter was then reused during evaluations, but we've now seperated them out. I think the following names might be best:
- train_batch_size_tokens
- store_batch_size_prompts
- eval_batch_size_prompts 
I generally like longer names if they are sufficiently clear but will admit these names are long and this could be annoying.

chanind:
That rename makes sense to me, I feel like that makes it more clear what's going on and reduces confusion around the difference between a "batch" of prompts vs tokens

tomMcGrath:
I like that rename - I just put a quick issue in to remind me to do the rename.

PR #132, opened 2024-05-09
tools to make tests more deterministic

bmillwood:
# Description

Pretty simple test change to add a pytest fixture that is run for every test, and use it to seed the RNG.

I don't strongly have evidence that not doing this caused test failures, or that doing it fixes them, but the principled case for doing this is pretty strong IMO and I can't think of any real downsides.

I think it's quite likely that there are other sources of non-determinism not fixed here (e.g. any non-global RNGs that are used in code), but incremental progress helps.

## Type of change

# Checklist:

    - none necessary
    - This is *possible* (e.g. by testing the output of `torch.get_rng_state`) but it's hard to imagine a version of this test that catches actual bugs.

### Performance Check.

- [ ] L0
- [ ] CE Loss
- [ ] MSE Loss

### DIFF
--- a/tests/unit/conftest.py
+++ b/tests/unit/conftest.py
@@ -1,6 +1,29 @@
+import random
+
+import numpy as np
 import pytest
+import torch
 
 from tests.unit.helpers import TINYSTORIES_MODEL, load_model_cached
+
+
+@pytest.fixture(autouse=True)
+def reproducibility():
+    """Apply various mechanisms to try to prevent nondeterminism in test runs."""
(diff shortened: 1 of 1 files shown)


jbloomAus:
Seems good! Will merge once tests pass.

jbloomAus:
@bmillwood, I think @chanind turned this off since it was causing errors. Wanted to check if this is the optimal choice (you two should chat).

bmillwood:
Looks like they didn't turn the whole thing off, just a small aspect of it. Have commented on the PR.

PR #134, opened 2024-05-09
Hooked toy model

### LINKED ISSUE #85: Add toy models architecture built on HookedRootModule
Per discussion in slack, SAELens has stale code for toy models. It would be great to bring that code back to life.

Toy models should be built on top of a class that itself is a subclass of HookedRootModule from transformer_lens, so that toy models can be run_with_cache and code for toy models and language models actually looks pretty similar.

For starters, the Anthropic ReLU output model is the only toy model we need working, but keeping generality for other types of models to be included in future PRs would be great!

Note: I'm going to be working on this this week (week of Apr 15), and wanted to make sure that it was visible that this is being thought about and worked on.

evanhanders:
# Description

Re-implements toy models as child classes of HookedRootModule. Most toy model functionality now lives in an abstract class, which specific toy models inherit from. Also adds some basic unit tests to toy models. Two toy models currently live in the toy model file:

- Anthropic's ReLU output model
- A model which trains using cross-entropy loss on a classification task with the same architecture as the ReLU output model

Fixes #85  (though there's still more toy model work to be done integrating with runners)

## Type of change

- [ x ] New feature (non-breaking change which adds functionality)

# Checklist:

- [ x ] I have commented my code, particularly in hard-to-understand areas
- [ I'm not sure how to do this ] I have made corresponding changes to the documentation
- [ I think they don't? ] My changes generate no new warnings 
- [ x ] I have added tests that prove my fix is effective or that my feature works
- [ x ] New and existing unit tests pass locally with my changes
- [ x ] I have not rewritten tests relating to key interfaces which would affect backward compatibility

- [ x ] I have run `make check-ci` to check format and linting. (you can run `make format` to format code if needed.)

### DIFF
--- a/sae_lens/training/toy_models.py
+++ b/sae_lens/training/toy_models.py
@@ -38,11 +38,21 @@ def cosine_decay_lr(step: int, steps: int):
     return np.cos(0.5 * np.pi * step / (steps - 1))
 
 
+def _init_importance(
+    importance: Optional[Union[float, Tensor]],
+    n_features: int,
+    device: str | torch.device,
+) -> Tensor:
+    if importance is None:
+        importance = t.ones(())
+    if isinstance(importance, float):
+        importance = t.tensor(importance)
+    assert isinstance(importance, Tensor)  # pyright can't seem to infer this
+    return importance.to(device).broadcast_to((n_features))
+
+
 @dataclass
 class ToyConfig:
-    # We optimize n_instances models in a single training loop to let us sweep over
-    # sparsity or importance curves  efficiently. You should treat `n_instances` as
-    # kinda like a batch dimension, but one which is built into our training setup.
     n_features: int = 5
     n_hidden: int = 2
     n_correlated_pairs: int = 0
@@ -69,16 +79,10 @@ class HookedToyModel(HookedRootModule, ABC):
             (cfg.n_features)
         )
 
-        importance = cfg.importance
-        if importance is None:
-            importance = t.ones(())
-        if isinstance(importance, float):
-            importance = t.tensor(importance)
-        assert isinstance(importance, Tensor)  # pyright can't seem to infer this
-        self.importance = importance.to(device).broadcast_to((cfg.n_features))
+        self.importance = _init_importance(cfg.importance, cfg.n_features, device)
 
     @abstractmethod
-    def forward(self, features: Tensor, return_type: str | None = None) -> Tensor:
+    def forward(self, features: Tensor, return_type: str) -> Tensor:
         """Forward pass, to be implemented by subclasses"""
 
     @abstractmethod
@@ -241,7 +245,7 @@ class ReluOutputModel(HookedToyModel):
     def forward(
         self,
         features: Float[Tensor, "... features"],
-        return_type: str | None = None,
+        return_type: str = "reconstruction",
     ) -> Float[Tensor, "... features"]:
         hidden = self.hook_hidden(
             einops.einsum(
@@ -261,8 +265,10 @@ class ReluOutputModel(HookedToyModel):
 
         if return_type == "loss":
             return self.calculate_loss(reconstructed, features)
-        else:
+        elif return_type == "reconstruction":
             return reconstructed
+        else:
+            raise ValueError(f"Unknown return type: {return_type}")
 
     def calculate_loss(
         self,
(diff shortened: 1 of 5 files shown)


evanhanders:
@jbloomAus & @chanind  -- I think the lines that aren't covered by tests are the plotting functions at the bottom of the toy model script. I had to modify them a bit when I removed the n_instances dimension from the toy models. Any thoughts on how or whether it's necessary to test those?

chanind:
The previous test coverage of these files was 0, so getting near 60% is a massive improvement already IMO, no need to go all the way unless you want to (always welcome of course!). The current coverage of this file is shown in [Codecov](https://app.codecov.io/gh/jbloomAus/SAELens/pull/134?src=pr&el=tree&filepath=sae_lens%2Ftraining%2Ftoy_models.py&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=Joseph+Bloom#diff-c2FlX2xlbnMvdHJhaW5pbmcvdG95X21vZGVscy5weQ==)

sae_lens/training/toy_models.py
```diff
@@ -271,17 +211,138 @@ def optimize(
...
+        """
+        Calculates the loss for a given batch, using this loss described in the Toy Models paper:
+
+            https://transformer-circuits.pub/2022/toy_model/index.html#demonstrating-setup-loss
+
+        Note, `model.importance` is guaranteed to broadcast with the shape of `out` and `batch`.
+        """
+        error = self.importance * ((batch - out) ** 2)
+        loss = einops.reduce(error, "batch features -> ()", "mean").sum()
+        return loss
```
chanind, line 281: This loss calculation would be good to cover in a test, but this is optional. You're already increasing test coverage a lot with this PR. I'm also happy to add tests to this after merging.

sae_lens/training/toy_models.py
```diff
@@ -271,17 +211,138 @@ def optimize(
...
+
+        # Add and setup hookpoints.
+        self.hook_hidden = HookPoint()
+        self.hook_out_prebias = HookPoint()
+        self.setup()
+
+    def forward(
+        self,
+        features: Float[Tensor, "... features"],
+        return_type: str | None = None,
```
chanind, line 244: nit: this feels odd to have a return type set to `None`, since this function does return reconstructions. Maybe the default could be "reconstruction" or something similar?

sae_lens/training/toy_models.py
```diff
@@ -271,17 +211,138 @@ def optimize(
...
+        is a one-hot encoding of the feature with the largest magnitude activation in the input.
+        Model outputs (raw logits) are weighted by importance before being passed through CE loss.
+
+        Note, `model.importance` is guaranteed to broadcast with the shape of `out` and `batch`.
+        """
+        max_feat_indices = t.argmax(batch, dim=-1)
+        loss = F.cross_entropy(
+            (self.importance * out).squeeze(), max_feat_indices.squeeze()
+        )
+        return loss
```
chanind, line 337: Here as well, this seems like an import part of the code to have coverage on. But only if you have time - I'm happy to add a test too

chanind (commented):
Very minor nits, overall this is a definite improvement to the code. I thought there was going to be a refactoring of the `ActivationsStore` to allow the Toy Model SAEs to be trained via the same code path as LM SAEs - is that part of a future PR?

The toy model training could use some simple tests, things like, e.g., setting `n_hidden = n_features` for a small model and assert that it can get close to perfect reconstruction (it should be able to do that if there's no actual squeeze). I don't think that should hold up merging this though, I'm happy to add those tests.

evanhanders:
Awesome, thanks for the feedback @chanind ! I'm out of the office today but will make these changes and get things ready to merge on Monday.

And yeah, the changes to ActivationStore, etc. that we talked about will come in a separate PR. Wanted to keep things a bit more segmented so we could do more thorough reviews of (a) toy model code and (b) SAE training on toy models code.

chanind:
I had some time today so just went and prepped this for merging. Figure it's best to get it merged quickly to avoid merge conflicts later

sae_lens/training/toy_models.py
```diff
@@ -271,17 +215,141 @@ def optimize(
...
+        extra_feature_value: float = 1e-6,
+    ):
+        super().__init__(cfg)
+        self.extra_feature_value = extra_feature_value
+
+        self.W = nn.Parameter(
+            nn.init.xavier_normal_(t.empty((cfg.n_hidden, cfg.n_features + 1)))
+        )
+        self.b_final = nn.Parameter(t.zeros((cfg.n_features + 1)))
+        self.importance = _init_importance(cfg.importance, cfg.n_features + 1, device)
```
chanind, line 316: @evanhanders I had to tweak this to get training tests to run, since otherwise `self.importance` has 1 less dim than the batch

evanhanders: Ah! Good catch -- thanks for fixing this; looks right! I'm not entirely logically sure what 'importance' should be implemented as in the CE model so this seems like the right first step.

PR #139, opened 2024-05-12
Remove cuda cache emptying in evals.py

tomMcGrath:
# Description

Currently evals.py includes a call to `torch.cuda.empty_cache()`. This causes inconsistent memory usage across runs, which makes OOMs unpredictable and might cause extra recompilation (though I'm not sure on this). `torch.cuda.empty_cache` doesn't actually free up additional memory for pytorch - it simply releases it from the pytorch allocator (I was surprised by this, discussion [here](https://pytorch.org/docs/stable/notes/cuda.html#memory-management) and [here](https://discuss.pytorch.org/t/about-torch-cuda-empty-cache/34232)). It's also not recommended for use by end-users except in edge cases, so I propose we remove it.

I've tested this change with and without compilation on GPT-2-S SAEs. All unit tests continue to pass.

## Type of change

# Checklist:

- [x] I have commented my code, particularly in hard-to-understand areas (n/a: code removal)
- [] I have added tests that prove my fix is effective or that my feature works

### Performance Check.

- [ ] L0
- [ ] CE Loss
- [ ] MSE Loss

### DIFF
--- a/sae_lens/training/evals.py
+++ b/sae_lens/training/evals.py
@@ -62,9 +62,6 @@ def run_evals(
     sae_out, _, _, _, _, _ = sparse_autoencoder(original_act)
     del cache
 
-    if "cuda" in str(model.cfg.device):
-        torch.cuda.empty_cache()
-
     l2_norm_in = torch.norm(original_act, dim=-1)
     l2_norm_out = torch.norm(sae_out, dim=-1)
     l2_norm_in_for_div = l2_norm_in.clone()


jbloomAus:
Huh, weird! I didn't know this. Thanks for catching that.

jbloomAus:
Accepting when builds pass.

tomMcGrath:
Yeah it's a bit weird! I'm sending over a summary about VRAM usage in a min

hijohnnylin:
sorry for my lack of knowledge on this, does this mean that we should remove all usages of empty_cache since it's not recommended for end users?
for example in these scripts:
https://github.com/jbloomAus/SAELens/blob/bdef2cf0e4ad3e3070f9be1c46a1adf13094eb13/scripts/caching_replication_how_train_saes.py#L52-L55

PR #143, opened 2024-05-13
Fix checkpointing of training state that includes a compiled SAE

tomMcGrath:
# Description

Compiled SAEs currently break checkpointing. The root cause of this is that compiled models cannot be saved with torch.save(compiled_model), but should instead be saved using their state dict (see [PyTorch ref](https://pytorch.org/get-started/pytorch-2.0/#serialization)). 

The model checkpointing code serializes the compiled model using its state_dict, but `SAETrainContext.save` does not. The issue comes from `L1Scheduler` having a reference to the SAE, which has been compiled. When `SAETrainContext.save` is called, because `L1Scheduler` has no state_dict, the code attempts to save a compiled model, leading to a crash.

This PR adds a state dict to L1Scheduler, as well as restoring the L1Scheduler from that state dict, which now allows for proper serialisation.

## Type of change

# Checklist:

Having some issues with `make test` but have run `make format` and run tests inside VSCode.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,23 @@
 
 
 
+## v2.1.2 (2024-05-13)
+
+### Fix
+
+* fix: Fix issues with resumption testing (#144)
+
+* fix always-true comparison in train context testing
+
+* set default warmup steps to zero
+
+* remove unused type attribute from L1Scheduler
+
+* update training tests to use real context builder
+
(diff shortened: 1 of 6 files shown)


jbloomAus:
My guess is this will work after a rebase?

tomMcGrath:
yeah, looks like it's all good now

jbloomAus:
Thanks!

PR #148, opened 2024-05-14
Pretokenize runner

### LINKED ISSUE #34: Create our own implementation for uploading pretokenized datasets to Huggingface.
Alan Cooney implemented this [here](https://github.com/ai-safety-foundation/sparse_autoencoder/blob/main/docs/content/pre-process-datasets.ipynb). We should reference his code, but it would be good to have our own implementation here for convenience.

chanind:
# Description

This PR adds a pretokenize runner which can be used to pre-tokenize / chunk / shuffle datasets and upload them to Huggingface. I plan on adding a follow-up PR in the future to harmonize the `ActivationsStore` to use the same logic for generating batches as is used here, but decided to hold off on that for now to avoid merge conflicts and to keep this PR more easily reviewable.

This PR has several design considerations after discussion in the OSMI slack:
- Support specifying a special token to use only at the beginning of every batch
- Support specifying a special token to use at the begging of each new sequence
- Support specifying a special token to use as a separator between sequences
- It should be possible for these tokens to occur one after another if desired (e.g. `seq 1 <eos> <bos> seq 2`)

The main batching function, `concat_and_batch_sequences()`, is written as a Python Generator, and thus is an iterator, so it can be used directly inside of `ActivationsStore` in the future to generate batches of tokens.

This PR also uploads a `sae_lens.json` metadata file along with the dataset that specifies info about how the dataset was generated, what version of SAELens was used to generate it, special tokens used, context_size, etc... In the future, we can read this data in when loading a pretokenized dataset in `ActivationsStore`.

This PR also includes a tutorial notebook on how to use this runner.

**NOTE**
The special tokens config for this runner differs from the way our other configs work, as our current configs only have a `prepend_bos` token option, but not the option to customize this behavior to the extent in this PR. I'm not sure if the names I used for these special token params is ideal, but definitely open to feedback on this! Longer term, we should use the same special token options for the other configs as well.

A sample pretokenized dataset generated with this script can be found [here](https://huggingface.co/datasets/chanind/c4-10k-tokenized-gpt2), and the corresponding metadata json can be found [here](https://huggingface.co/datasets/chanind/c4-10k-tokenized-gpt2/blob/main/sae_lens.json)

Fixes #34 

## Type of change

# Checklist:

### DIFF
--- a/sae_lens/__init__.py
+++ b/sae_lens/__init__.py
@@ -2,9 +2,14 @@ __version__ = "2.1.2"
 
 from .training.activations_store import ActivationsStore
 from .training.cache_activations_runner import CacheActivationsRunner
-from .training.config import CacheActivationsRunnerConfig, LanguageModelSAERunnerConfig
+from .training.config import (
+    CacheActivationsRunnerConfig,
+    LanguageModelSAERunnerConfig,
+    PretokenizeRunnerConfig,
+)
 from .training.evals import run_evals
 from .training.lm_runner import language_model_sae_runner
+from .training.pretokenize_runner import pretokenize_runner
 from .training.sae_group import SparseAutoencoderDictionary
 from .training.session_loader import LMSparseAutoencoderSessionloader
 from .training.sparse_autoencoder import SparseAutoencoder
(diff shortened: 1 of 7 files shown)


jbloomAus:
This looks great. Looking forward to playing with this. 

Some minor notes:
1. I think we're getting to the point where runners should be brought a level up in order to make it clear what the critical interfaces for the codebase are.
2. I think we should make a diagram to show people which scripts to run in which order to use the whole pipeline. 

(not critical now though).

A good follow up PR:
- Make a list of datasets people can use for common models (pretokenized and not pretokenized)
- Add to that a list of datasets people have generated and shared with this script. 

I'm going to try to make some Gemma-2b and Gemma-7b tokenized datasets with this soon.

PR #150, opened 2024-05-15
Fix normalisation

tomMcGrath:
# Description

Anthropic's SAE training code normalises activations before they enter the SAE so that their average L2 norm is sqrt(d_model). Although we have normalisation activation in the code, it is not currently applied during evals which both harms the cross-entropy loss and makes hyperparameter transfer across layers impossible as different layers have substantially different mean norms.

This PR changes estimation of the scale factor to occur once, with a larger sample size, before SAE training, and modifies the eval code to up- and downscale activations as part of the hook function as well as correctly estimate the L2 norm of activations that the SAE will actually see during training.

The code in this PR isn't the prettiest and I aim to create a tidier version next week but we need this functionality. I've marked areas for tidying with `TODO(tomMcGrath)` and will open an issue to track this. EDIT: this is now tracked in issue #151.

## Type of change

# Checklist:

### Performance Check.

- [ ] L0
- [x] CE Loss
- [ ] MSE Loss

[This W&B dashboard ](https://wandb.ai/tmcgrath/gpt-2-sweep-15may24-try-normalisation/workspace.)shows the effect of the fix:
- `metrics/l2_norm_in` shows the L2 norm of the inputs fluctuating around sqrt(d_model)
- `metrics/CE_loss_score` shows the effects of correctly up- and downsampling, as well as the effects of no normalisation (the state pre-fix) and only upscaling

### DIFF
--- a/sae_lens/training/activations_store.py
+++ b/sae_lens/training/activations_store.py
@@ -122,6 +122,8 @@ class ActivationsStore:
         self.n_dataset_processed = 0
         self.iterable_dataset = iter(self.dataset)
 
+        self.estimated_norm_scaling_factor = 1.
+
         # Check if dataset is tokenized
         dataset_sample = next(self.iterable_dataset)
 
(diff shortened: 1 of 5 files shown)


jbloomAus:
Awesome work! Thanks so much!

PR #159, opened 2024-05-22
Move activation store to cpu

tomMcGrath:
# Description

Currently the ActivationStore class stores the activation buffer (which is typically large) in the same location as the model; this location is normally VRAM which is a scarce resource. 

This PR allows the ActivationStore to store the activation buffer on a different device, including CPU. The device to use is specified by the new `act_store_device` config entry, which defaults to using the same device as the model to ensure backwards compatibility.

The linked W&B dashboard below demonstrates that this change has no effect on throughput vs storing activations in VRAM for a typical workload and allows at least an 8x scaleup of activation buffer size.

## Type of change

# Checklist:

- [] I have made corresponding changes to the documentation

### Performance Check.

- [x] L0
- [x] CE Loss
- [x] MSE Loss

This is a performance improvement: increasing metrics is not the goal. This PR maintains throughput while freeing up VRAM: see [W&B dashboard](https://wandb.ai/tmcgrath/gpt-2-sweep-21may24-acts-to-cpu/workspace).


jbloomAus:
Seems good. Do you have a moment to write a few tests? Bit nervous we could have a bug and not realise it here.

tomMcGrath:
Sure, will add some to this pr

tomMcGrath:
I've added some tests and fixed an issue in the test cacher + an issue the new tests exposed in my code.

jbloomAus:
Awesome work! Thanks!

PR #178, opened 2024-06-10
Adding Mistral SAEs

JoshEngels:
# Description

This is a brief PR that adds support for loading the Mistral 7B SAEs we trained for https://arxiv.org/abs/2405.14860. As an important note, we normalized the activations before the SAE forward pass and unnormalized them afterwards like the Anthropic February update. I tried adding this to the library but it was messy because encode and decode are sometimes called individually during evaluation, so we can't just normalize and un-normalize in the forward pass, so for now I was just thinking it would be on the user to do correctly. Let me know if there is a better way I should do this!

I also evaluated these SAEs with code here: https://github.com/JoshEngels/MultiDimensionalFeatures/blob/main/sae_multid_feature_discovery/upload_to_huggingface.py
Happy to upload this code to the tutorial repo as well since it's useful to see how to do the normalization, but didn't want to make a big mess in this PR.

SAE statistics on first few hundred documents of the pile from this evaluation:

| Layer | Variance Explained | L0 | % CE Loss Recovered |
| -------- | ------- | ------- | ------- |
| 8 | 74% | 81.8 | 99.6% |
| 16 | 85% | 73.7 | 99.2% |
| 24 | 72.13% | 75.2 | 98.1% |

The % CE loss recovered is pretty high even though the variance explained is low; I'm a bit skeptical of this, but I don't see a bug in my evaluation code (it's mostly copied from the existing metrics), and I've also seen similar things when I trained SAEs before.

## Type of change

# Checklist:

(formatting passed but ran into some problems with the other tests, probably unrelated to my PR)
### Performance Check.

- [ ] L0
- [ ] CE Loss
- [ ] MSE Loss


nix-apollo:
I think (someone correct me if I'm wrong) that SAE config has a `normalize_activations` flag. This doesn't change the forward pass of the SAE itself, but it is checked by e.g. [the evaluation code](https://github.com/jbloomAus/SAELens/blob/c4297ac2203d50dba7589b612a625ac38cc15a2b/sae_lens/evals.py#L149-L160) to know how to preprocess the activations.

You'd have to make that change in the huggingface repo.

jbloomAus:
This is a different normalization procedure than the one Anthropic suggests in the April update which normalizes all activations by some constant for the whole dataset. The procedure in this PR let's re-estimates the normalization factor per example. With the procedure we currently have, you can fold the normalization constant into the SAE encoder post training but with this method you can't. On the other hand, with the procedure you've used, we don't need to worry about estimating the constant all the time. I probably want to support both but we should have very clear flags and a condition that says you can't do both at once. 

Not sure what to call each. Does the difference make sense to you @JoshEngels ?

ianand:
@JoshEngels  @jbloomAus I noticed that in this PR the tutorial using_an_sae_as_a_steering_vector.ipynb has changed from using gpt2-small to gemma-2b as the model. Any particular reason? Should I expect it's fine to continue using with gpt2-small?

jbloomAus:
That was a mistake! I'll fix it.

PR #181, opened 2024-06-15
feat: harmize activation store and pretokenize runner

chanind:
# Description

This PR shares the tokenization / batching / concat logic between `ActivationStore` and `PretokenizeRunner`, so both places can have identical behavior, and reduce code duplication.

For pre-tokenizated datasets, this PR just uses the tokenized sequences directly without attempting to do any further reshaping of tokens. This is both a performance improvement, and tries to avoid reshaping already batched tokens as this will result in a lot of spurious partial sequences separated by BOS tokens. This means that for pretokenized datasets, the `context_size` must match the `context_size` used by the pre-tokenized dataset. If these don't match, an error is now thrown.

To avoid breaking changes and to keep the options as simple as possible, this PR keeps the `prepend_bos` option in SAE runner. This is interpreted as using `bos` as the first token in each batch, as well as being a separator token between concatenated sequences.

this PR leave the pretokenize runner with its current tokenization settings which allow more complicated tokenization / batching, e.g. using `eos` as a separator instead of `bos`. If users need more complicated tokenization/batching, they can achieve this with the pretokenize runner. If in the future we find there's a need for these more complicated tokenization settings in the SAE runner, we can make the breaking change to add those options at that time.

--

I added a benchmarking test for running `ActivationsStore.get_batch_tokens()` for 100 batches of size 50, with context_size 512 tokens. This tests compare pre-tokenization vs raw text, and I compared the results of the PR vs running the same benchmarking on main.

### Results for main
get_batch_tokens() duration with text dataset: 6.75s
get_batch_tokens() duration with pretokenized dataset: 0.38s

### Results for this PR

get_batch_tokens() duration with text dataset: 6.72s
get_batch_tokens() duration with pretokenized dataset: 0.24s

It looks like this PR doesn't have much impact on the performance of non-tokenized datasets, but is about 40% faster than the current main for pre-tokenized datasets.
 

## Type of change

- [x] Performance improvement and refactor

# Checklist:


jbloomAus:
@chanind Let's discuss on our call this week.

jbloomAus:
To do:
- [ ] Add a prompt when not pre-tokenizing indicating you should pre-tokenize (and how).
- [ ] Add to the readme / docs a list of currently pre-tokenized datasets (which models / datasets and if we used our pre-tokenizer runner, then what that config was).
- [ ] Add a check that the tokenizer was the same for the dataset. 
- [ ] Rebase. 
- [ ] Optional: Write an issue for generating / mixing in chat model data for SAEs training on chat models.

chanind:
I added a table of pre-tokenized datasets to the docs, screenshot below:

localhost_8000_training_saes_

chanind:
I'll address automatically checking for tokenizer compatibility in a follow-up PR

PR #183, opened 2024-06-15
[WIP] refactor: sae forward pass

### LINKED ISSUE #182: [Proposal] Add MLP transcoders
### Proposal 

Support training, loading, and inference of MLP transcoders.  

### Motivation

MLP transcoders were trained by Jacob Dunefsky and Philippe Chlenski and have been shown to be useful. 

### Pitch

- Implement a `HookedTranscoder` class analogous to `HookedSAE` and using similar functionality. 
- Implement a transcoder training runner. 
- Support loading pre-trained transcoder checkpoints.

### Checklist

- [X] I have checked that there is no similar [issue](https://github.com/jbloomAus/SAELens/issues) in the repo (**required**)

dtch1997:
# Description

Add support for training, loading, and running inference on MLP transcoders.

Fixes #182 

## Type of change

# Checklist:

### Performance Check.

- [ ] L0
- [ ] CE Loss
- [ ] MSE Loss

### DIFF
--- a/sae_lens/sae.py
+++ b/sae_lens/sae.py
@@ -216,43 +216,23 @@ class SAE(HookedRootModule):
         feature_acts = self.encode(x)
         sae_out = self.decode(feature_acts)
 
-        if self.use_error_term:
-            with torch.no_grad():
-                # Recompute everything without hooks to get true error term
-                # Otherwise, the output with error term will always equal input, even for causal interventions that affect x_reconstruct
-                # This is in a no_grad context to detach the error, so we can compute SAE feature gradients (eg for attribution patching). See A.3 in https://arxiv.org/pdf/2403.19647.pdf for more detail
-                # NOTE: we can't just use `sae_error = input - x_reconstruct.detach()` or something simpler, since this would mean intervening on features would mean ablating features still results in perfect reconstruction.
+        if not self.use_error_term:
+            return self.hook_sae_output(sae_out)
 
-                # move x to correct dtype
-                x = x.to(self.dtype)
-
-                # handle hook z reshaping if needed.
-                sae_in = self.reshape_fn_in(x)  # type: ignore
-
-                # handle run time activation normalization if needed
-                sae_in = self.run_time_activation_norm_fn_in(sae_in)
-
-                # apply b_dec_to_input if using that method.
-                sae_in_cent = sae_in - (self.b_dec * self.cfg.apply_b_dec_to_input)
-
-                # "... d_in, d_in d_sae -> ... d_sae",
-                hidden_pre = sae_in_cent @ self.W_enc + self.b_enc
-                feature_acts = self.activation_fn(hidden_pre)
-                x_reconstruct_clean = self.reshape_fn_out(
-                    self.apply_finetuning_scaling_factor(feature_acts) @ self.W_dec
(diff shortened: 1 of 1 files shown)


dtch1997:
Initial Wandb run here: https://wandb.ai/dtch1997/benchmark/workspace

Benchmarked using the following command: 
```
poetry run pytest tests/benchmark/test_language_model_transcoder_runner.py --profile-svg -s
```

dtch1997:
Current status of the PR: 
- We can train transcoders from scratch

The next priority might be to support pre-trained MLP transcoders.
Going to work on this tomorrow-ish

dtch1997:
Some notes on architecture. 
- minimal coupling, maximal duplication: Implement Transcoder as a totally separate class with code copy-pasted from SAE.
- maximal coupling, minimal duplication: Implement both Transcoder and SAE in the same class via a config option. (suggested by Jacob). 
- somewhere in the middle: inheritance / polymorphism, which is what I've started doing.

the right balance depends on whether we expect that we'll need to change Transcoder independently of SAE (if so, we'd prefer less coupling) and on whether relevant functionality can be reused between the two modules (if we can refactor certain things into external functions, or if it turns out that we can use the same Trainer class for both SAE and Transcoder, then we'll have a lot less duplication). 

Feedback on the last point above would be very useful.

sae_lens/sae.py
```diff
@@ -443,3 +456,41 @@ def tanh_relu(input: torch.Tensor) -> torch.Tensor:
...
+        super().__init__(cfg, use_error_term)
+
+    def initialize_weights_basic(self):
+        super().initialize_weights_basic()
+
+        # NOTE: Transcoders have an additional b_dec_out parameter.
+        # Reference: https://github.com/jacobdunefsky/transcoder_circuits/blob/7b44d870a5a301ef29eddfd77cb1f4dca854760a/sae_training/sparse_autoencoder.py#L93C1-L97C14
+        self.b_dec_out = nn.Parameter(
+            torch.zeros(self.cfg.d_out, dtype=self.dtype, device=self.device)
+        )
```
nix-apollo, line 487: I don't understand why the extra bias is needed. I'm probably just confused and missing something, but it would make the implementation simpler if you don't need it.

I understand that in normal SAEs people sometimes subtract b_dec from the input. This isn't really necessary but has a nice interpretation of choosing a new "0 point" which you can consider as the origin in the feature basis.

For transcoders this makes less sense. Since you aren't reconstructing the same activations you probably don't want to tie the pre-encoder bias with the post-decoder bias.

Thus, in the current implementation we do:
$$z = ReLU(W_{enc}(x - b_{dec}) + b_{enc})$$
and
$$out = W_{dec} x +b_\text{dec out}$$
This isn't any more expressive, you can always fold the first two biases ($b_{dec}$ and $b_{enc}$ above) into a single bias term. I don't see a good reason why it would result in a more interpretable zero point for the encoder basis either.

Overall I'd recommend dropping the complexity here, which maybe means you can just eliminate the Transcoder class entirely.

dtch1997: this makes sense! i'll try dropping the extra `b_dec` term when training. I was initially concerned about supporting the  previously-trained checkpoints, but as you say weight folding should solve that.

PR #187, opened 2024-06-17
fix: allow settings trust_remote_code for new huggingface version

chanind:
# Description

The newest version of Huggingface transformers won't allow loading old datasets that require allowing code execution without explicitly setting `trust_remote_code=True`. As a result, tests are now failing on CI since it loads the most recent version of transformers. This PR adds an option `dataset_trust_remote_code` which the user can set to allow this remote code execution so old datasets can be loaded in new versions of transformers, and also fixes CI.

## Type of change

# Checklist:


jbloomAus:
@chanind Why make it optional None? We should just default to False?

chanind:
I was worried that defaulting to `False` would be a breaking change, since then previously working code would no longer work if it's using an old dataset that requires setting the `trust_remote_code` variable. Setting it to `None` lets this be breaking or not based on Huggingface and not us.

jbloomAus:
Ahhh I imagine setting to true avoids breaking change since if you trust
remote code it will run?

PR #188, opened 2024-06-17
feat: Support Gated-SAEs

curt-tigges:
# Description

- Provides support for training SAEs with a gated architecture per [this paper](https://arxiv.org/abs/2404.16014).
- Example run [here](https://wandb.ai/jbloom/gated_sae_benchmark_curts_branch/runs/528ks8h6)

To use gated SAEs simply add `architecture='gated'` to the training config which will default to standard otherwise. 

# Description

Draft PR for gated SAE with SFN loss. Currently, the actual training seems to have bad results, even when trained for a full hour. See https://wandb.ai/curt-tigges/gated_sae_testing?nw=nwusercurttigges, which includes a non-SFN run for comparison. The issue seems to be that even though SFN loss does indeed achieve a low count of dead features, L0 remains too high even when the coefficient is set so high that MSE loss starts to suffer. Also, note the weird periodic fluctuations in various metrics.

Currently submitted to facilitate bug-checking. Additionally, one test does not pass--perhaps we should discuss what this test should do since the types of loss returned by the trainer can vary.

Fixes # (issue)

## Type of change

# Checklist:

### Performance Check.

- [ ] L0
- [ ] CE Loss
- [ ] MSE Loss


jbloomAus:
Yeah somethings definitely weird here. Will investigate today and fix.

sae_lens/sae.py
```diff
@@ -184,16 +192,54 @@ def initialize_weights_basic(self):
                 torch.ones(self.cfg.d_sae, dtype=self.dtype, device=self.device)
             )
 
+    def initialize_weights_gated(self):
+        # Initialize the weights and biases for the gated encoder
+        self.b_enc = nn.Parameter(
```
jbloomAus, line 197: b_enc isn't getting used? We have b_gate, b_mag.

sae_lens/sae.py
```diff
@@ -184,16 +192,54 @@ def initialize_weights_basic(self):
...
+        )
+
     # Basic Forward Pass Functionality.
     def forward(
         self,
         x: torch.Tensor,
     ) -> torch.Tensor:
-
-        feature_acts = self.encode(x)
+        feature_acts = self.encode_fn(x)
```
jbloomAus, line 238: Aesthetically I like encode and decode. Will think about whether the _fn style is really good.

sae_lens/sae.py
```diff
@@ -184,16 +192,54 @@ def initialize_weights_basic(self):
...
         sae_out = self.decode(feature_acts)
 
-        if self.use_error_term:
+        # TEMP
+        if self.use_error_term and self.cfg.architecture != "gated":
```
jbloomAus, line 242: TO DO

sae_lens/sae.py
```diff
@@ -224,11 +270,28 @@ def forward(
...
+        x = self.reshape_fn_in(x)
+        sae_in = self.hook_sae_input(x - self.b_dec)
+
+        # Gating path
+        gating_pre_activation = sae_in @ self.W_enc + self.b_gate
+        active_features = (gating_pre_activation > 0).float()
+
+        # Magnitude path with weight sharing
+        magnitude_pre_activation = sae_in @ (self.W_enc * self.r_mag.exp()) + self.b_mag
+        feature_magnitudes = self.activation_fn(magnitude_pre_activation)
```
jbloomAus, line 286: No use of hooks!

sae_lens/training/sae_trainer.py
```diff
@@ -287,6 +287,8 @@ def _build_train_step_log_dict(
             "losses/mse_loss": mse_loss,
             "losses/l1_loss": l1_loss
             / self.current_l1_coefficient,  # normalize by l1 coefficient
+            "losses/sfn_sparsity_loss": output.sfn_sparsity_loss,
```
jbloomAus, line 290: Not convinced we should have a different loss for SFN since for vanilla SAE we use the l1 loss. I think we could change what we call the l1 loss to sparsity loss or something though.

sae_lens/training/training_sae.py
```diff
@@ -139,6 +143,15 @@ def __init__(self, cfg: TrainingSAEConfig, use_error_term: bool = False):
...
         super().__init__(base_sae_cfg)
         self.cfg = cfg  # type: ignore
+
+        self.encode_with_hidden_pre_fn = (
+            self.encode_with_hidden_pre
+            if cfg.architecture != "gated"
+            else self.encode_with_hidden_pre_gated
+        )
+
+        self.check_cfg_compatibility()
```
jbloomAus, line 153: We're checking config compatibility elsewhere (in the training config init). I think I'd prefer to put this there.

sae_lens/training/training_sae.py
```diff
@@ -184,6 +204,38 @@ def encode_with_hidden_pre(
 
         return feature_acts, hidden_pre_noised
 
+    def encode_with_hidden_pre_gated(
```
jbloomAus, line 207: Once we get to the point we're writing different methods, I do wonder if we should just make different classes.

sae_lens/training/training_sae.py
```diff
@@ -225,17 +277,51 @@ def training_forward_pass(
...
+            #     * torch.sum(via_gate_feature_magnitudes, dim=-1).mean()
+            # )
+            
+            # Shared variables
+            sae_in_centered = sae_in - self.b_dec
+            pi_gate = sae_in_centered @ self.W_enc + self.b_gate
+            
+            # SFN sparsity loss - summed over the feature dimension and averaged over the batch
+            sfn_sparsity_loss = current_l1_coefficient * torch.sum(
+                torch.abs(pi_gate) * self.W_dec.norm(dim=1), dim=-1
```
jbloomAus, line 298: This is incorrect, we should apply the Relu then use pi_gate_act here. Fixing this results in working benchmark runs with/without SFN.

tutorials/training_a_gated_sae.ipynb
jbloomAus, line 1: Don't need a special file for gated SAE training.

jbloomAus (commented):
@curt-tigges The PR as it stood was basically good except for 2 bugs in the training SAE code (always subtracting b_dec regardless of the config) and using the Pre-Relu activations in the SFN loss calculation. Once fixed I produced good runs with both. I'm going to finish this PR and merge, thanks for your help though!

PR #189, opened 2024-06-17
 Performance improvements + using multiple GPUs. 

jbloomAus:
# Description

I'm just working with some larger models and even though IO is still likely our bottleneck, there were a few easy wins with multiple GPUs I wanted to get out. 

These were:
- `ActivationsStore` was running code without `torch.no_grad()`. If the model was in eval mode this wouldn't be an issue but I've added these contexts to the key functions just in case. 
- `load_model` utility will now check if n_devices is set and force the model device to be "cuda". This is important for enabling the SAE to be on a device like "cuda:3" while the model is on devices "cuda:0" - "cuda:2". 
- As a result of wanting to facilitate the above, I have updated the trainer / evals modules so data is moved between devices as needed. 

We should add these to the docs at some point but here's an example of using multiple devices for the model, SAE and activations store. 

```python
import torch

from sae_lens.config import LanguageModelSAERunnerConfig
from sae_lens.sae_training_runner import SAETrainingRunner

# For brevity, assume some basic config dict
cfg_dict = BASE_CFG # assume you have a config with Gemma-2b 
# we have 4 devices in this example
assert torch.cuda.device_count() == 4

# Model goes on cuda:0 and cuda:1. Setting n_devices will cause us to assume we're on cuda. 
cfg_dict["model_from_pretrained_kwargs"] = {"n_devices": torch.cuda.device_count() - 2}  

# SAE goes on cuda:2
cfg_dict["device"] = "cuda:2"

# Activation store goes on cuda:3  (faster than having activation store on CPU
cfg_dict["act_store_device"] = "cuda:3"

# Instantiate Config.
cfg = LanguageModelSAERunnerConfig(**cfg_dict)  # type: ignore
# look at the next cell to see some instruction for what to do while this is running.
sae = SAETrainingRunner(cfg).run()
```

## Type of change

# Checklist:


kaf7179-svg:
@jbloomAus Hi thanks for the post and for the library. I was wondering if the above description is still the best way to do SAE training with multiple devices available. When I tried the above setup (2 for model, 1 for SAE, 1 for activation store) I got this error:

>   File "/venv/lib/python3.12/site-packages/transformer_lens/components/layer_norm.py", line 57, in forward
    return self.hook_normalized(x * self.w + self.b).to(self.cfg.dtype)
                                ~~^~~~~~~~
RuntimeError: Expected all tensors to be on the same device, but found at least two devices, cuda:0 and cuda:1!

Maybe things have changed since the above update? Thanks!

PR #198, opened 2024-06-29
Added markdown version of pretrained_saes.yaml

skunnavakkam:
# Description

This adds a markdown file that converts the yaml file in pretrained_saes.yaml to a nice table in the docs. This was referenced in index.md

No code was changed.

### DIFF
--- a/docs/importable_saes.md
+++ b/docs/importable_saes.md
@@ -1,0 +1,129 @@
+
+# Importable SAEs
+
+This is a list of  SAEs importable from the SAELens package. Click on each link for more details.
+- [Importable SAEs](#importable-saes)
+  - [gpt2-small-res-jb](#gpt2-small-res-jb)
+  - [gpt2-small-hook-z-kk](#gpt2-small-hook-z-kk)
+  - [gpt2-small-mlp-tm](#gpt2-small-mlp-tm)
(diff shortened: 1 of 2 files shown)


chanind:
Was this generated via a script? If so it would be great to commit that script as well, and maybe add it as a target in `Makefile` so we can run something like `make update-pretrained-saes-docs` to keep this up-to-date in the future as more SAEs are added to the YAML

skunnavakkam:
I did the conversion with Claude and then combed over it to make sure no errors, but I think that scripting shouldn't be too hard and I don't mind implementing

jbloomAus:
Scripting would be good. I'd prefer not to quote those L0's / variance explained since they were on a random piece of text for a stop gap testing solution. Will set up benchmarking soon (after I accept Josh's PR) and will report the stats coming out of there. Thanks for this!

jbloomAus:
I've made a script that does this now. https://jbloomaus.github.io/SAELens/sae_table/

Would love any feedback on the page / suggestions! My current vibe is:
- Maybe one monolithic table would be nicer?
- We should probably add comments for each SAE to explain which to use.
- Once we have SAE Benchmarking running across all of these we should have a toggle for "show me config in detail" vs "show me the summary stats" in detail.

PR #206, opened 2024-07-02
SAETrainingRunner takes optional HFDataset

Lewington-pitsos:
Allows users to pass in datasets which are not on huggingface when training using the `SAETrainingRunner`

Code of the following kind (which is quite useful imo, especially when testing) now runs:

```python

... 

dataset: IterableDataset = cast(IterableDataset, load_dataset(
    "apollo-research/roneneldan-TinyStories-tokenizer-gpt2",
    split="train",
    streaming=True,
    trust_remote_code=True,
))

dataset = dataset.take(100)

sparse_autoencoder = SAETrainingRunner(cfg, dataset).run()
```

The changes are based on an existing pattern for passing in non-path datasets to the `ActivationStore`

Also added a more explicit error for when the activation buffer size is too large for the dataset passed in.

IMO we should give users the option of passing in `None` to the `dataset_path` across all kinds of config for cases like this as it will lead to less confusion, but I am not familiar enough with the codebase to make such an important change to so many interfaces. 

Another reasonable change would be to make the default `dataset_path` an empty string instead of good old `NeelNanda/c4-tokenized-2b` since users should probably always choose which dataset they are using explicitly.

Very open to feedback on this PR

## Type of change

# Checklist:

### DIFF
--- a/sae_lens/config.py
+++ b/sae_lens/config.py
@@ -5,6 +5,7 @@ from typing import Any, Literal, Optional, cast
 
 import torch
 import wandb
+from datasets import Dataset, DatasetDict, IterableDataset, IterableDatasetDict
 
 from sae_lens import __version__
 
@@ -18,6 +19,8 @@ DTYPE_MAP = {
     "torch.float16": torch.float16,
     "torch.bfloat16": torch.bfloat16,
 }
+
+HfDataset = DatasetDict | Dataset | IterableDatasetDict | IterableDataset
 
 
 @dataclass
--- a/sae_lens/config.py
+++ b/sae_lens/config.py
@@ -116,7 +116,7 @@ class LanguageModelSAERunnerConfig:
     hook_eval: str = "NOT_IN_USE"
     hook_layer: int = 0
     hook_head_index: Optional[int] = None
-    dataset_path: str = "NeelNanda/c4-tokenized-2b"
+    dataset_path: str = ""
     dataset_trust_remote_code: bool = True
     streaming: bool = True
     is_dataset_tokenized: bool = True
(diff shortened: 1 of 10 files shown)


jbloomAus:
I'm conflicted. On the one hand, I want to make it so the SAE training runner cfg is serializable (for reproducibility) and this side steps that (setting a dangerous precedent) but on the other hand, clearly we need more flexibility here. We should probably have this + the model equivalent so people can use transformer_lens models not loadable as pretrained. 

I think I'm happy with this conditional on:
- [ ] rename dataset to override_dataset or something like that. 
- [ ] log / print a warning that says "this training run won't be reproducible via config alone because of use of this parameter. 

On changing the default. Yeah this is probably fine. We probably shouldn't have defaults.

@Lewington-pitsos would you be interested in making another PR so people can provide a model like this as well? So many people have asked for this.

Lewington-pitsos:
- default `dataset_path` replaced with empty string
- dataset -> override_dataset 
- log warning whenever you pass in `override_dataset`

done, done, done. The model thing sounds like a good change too, I can chuck it on my todo...

having users specify a thing (dataset, model etc) and then also override it is a bit sus though. Maybe the long term solution is a `LanguageModelSAERunnerConfig` and a `NonReproducableLanguageModelSAERunnerConfig` and the user just **explicitly** chooses the one they want and suffers the consequences

Lewington-pitsos:
I'm removing the `ValueError` pertaining to the `StopIteration` which I just added because it looks like there might be something else odd going on there and mixing `ValueError` and `StopIteration` muddies the water further.

jbloomAus:
Thanks! Really appreciate this!

@Lewington-pitsos  I really liked your Medium article btw. 

>Maybe the long term solution is a LanguageModelSAERunnerConfig and a NonReproducableLanguageModelSAERunnerConfig and the user just explicitly chooses the one they want and suffers the consequences

Ahh this feels over complicated for now. I think we're good as is.

Lewington-pitsos:
> really liked your Medium article btw.

Thanks mate! Very good to hear I'm not barking up _entirely_ the wrong tree haha, super new to the space.

There must be a really clean way to handle this. I wonder what Soumith Chintala would do...

Lewington-pitsos:
By the way, I'm vibing the strict typing and amount of unit testing in this repo super hard   🙏

Issue #209, opened 2024-07-02
[Proposal] Make SAE Lens compatible with the latest Transformers / TransformerLens versions


jbloomAus:
### Proposal 

Update T-lens / transformers and make sure tests pass. 

### Motivation

I found a few tests failed when I installed transformerlens from source with the latest transformers version (which seemed to be needed to avoid other errors. 

### Checklist

- [x] I have checked that there is no similar [issue](https://github.com/jbloomAus/SAELens/issues) in the repo (**required**)

PR #214, opened 2024-07-05
Updated pyproject.toml to use TL ^2.0, and to use fork of sae-vis tha…

curt-tigges:
…t also uses TL ^2.0

# Description

Addresses https://github.com/jbloomAus/SAELens/issues/209 by updating to use TransformerLens ^2.0.0. In order to do this, I also needed to fork and modify sae-lens since it is not compatible with TL ^2.0.0. Thus, the changes rely on my current fork, but I propose making a PR to sae-vis so that we can go back to relying directly on the main library.

Fixes # (issue)

## Type of change

- [X] Bug fix (non-breaking change which fixes an issue)

# Checklist:


chanind:
I don't think we publish the library to PyPI while depending on a module on git. Can you open a PR to sae-vis with your change so it can also publish the TL2.0 version to PyPI?

jbloomAus:
I think we should just depend on SAE Dashboard instead of SAE Vis. https://github.com/jbloomAus/SAEDashboard

Since we're planning to move over to that anyway. This seems like a reason for us to make sure SAE Dashboard is documented well enough and to make tutorials in SAE Lens refer to it instead of SAE Vis.

jbloomAus:
Edit: @curt-tigges I'd rather SAE Lens not depend on SAE Dashboard at all! SAE Dashboard will need to depend on SAE Lens and I'd like to avoid circular dependencies.

curt-tigges:
I did a brief check and didn't see any SAELens code using sae-vis or SAEDashboard directly, but I can do a more careful check to make sure and remove the dependencies.

curt-tigges:
Updated to remove reliance on sae-vis and SAEDashboard. Notebooks have install commands included.

In order to make the tests pass with TL ^2.0.0, I added an error handler for the `get_activations` function, but this might be an issue that TransformerLens should address (the view command failed, but reshape worked).

PR #216, opened 2024-07-07
Memmap shuffling 

Lewington-pitsos:
# Description

A fork of the `improved_io` branch https://github.com/jbloomAus/SAELens/tree/improved_io

- [x] fix `test_cache_activations_runner_saving` (a memmap was being initialized as `float32` and then read as `float16` causing the loaded memmap to seem twice as large as expected
- [x] fix all existing typing errors across `ActivationsStore` and `CacheActivationsRunner`
- [x] make `ActivationsStore.get_batch` use the new memmap strategy instead of the old dataloader
- [ ] get all prior unit tests passing after this change
- [ ] bring in line with main
- [ ] add some new tests to cover the new `get_batch` functionality

## Type of change

### Performance Check.

- [ ] L0
- [ ] CE Loss
- [ ] MSE Loss

### DIFF
--- a/sae_lens/training/activations_store.py
+++ b/sae_lens/training/activations_store.py
@@ -420,7 +420,7 @@ class ActivationsStore:
         return stacked_activations
 
     @torch.no_grad()
-    def get_buffer(self) -> np.ndarray[Any, np.dtype[Any]]:
+    def get_buffer(self) -> np.memmap[Any, np.dtype[Any]]:
         context_size = self.context_size
         batch_size = self.store_batch_size_prompts
         d_in = self.d_in
@@ -455,8 +455,6 @@ class ActivationsStore:
                 shape=(total_size, context_size, num_layers, d_in),
             )
 
-            print('memmapped file shape', new_buffer.shape)
-
             for refill_batch_idx_start in refill_iterator:
                 # move batch toks to gpu for model
                 refill_batch_tokens = self.get_batch_tokens().to(self.model.cfg.device)
@@ -477,7 +475,7 @@ class ActivationsStore:
 
         return new_buffer
 
-    def save_buffer(self, buffer: np.memmap[Any, Any], path: str):
+    def save_buffer(self, buffer: np.memmap[Any, np.dtype[Any]], path: str):
         # If buffer is already a memmap, we just need to flush it
         if buffer.filename != path:
             # If the paths are different, we need to create a new memmap
@@ -490,7 +488,7 @@ class ActivationsStore:
             # If the paths are the same, we just need to flush the existing buffer
             buffer.flush()
 
-    def load_buffer(self, path: str) -> np.ndarray[Any, np.dtype[Any]]:
+    def load_buffer(self, path: str) -> np.memmap[Any, np.dtype[Any]]:
         # Load the memory-mapped array
         memmap_file = np.memmap(path, dtype=self.numpy_dtype, mode="r")
         memmap_file = memmap_file.reshape(-1, 1, self.d_in)
--- a/sae_lens/training/activations_store.py
+++ b/sae_lens/training/activations_store.py
@@ -475,7 +475,8 @@ class ActivationsStore:
 
         return new_buffer
 
-    def save_buffer(self, buffer: np.memmap[Any, np.dtype[Any]], path: str):
+    @classmethod
+    def _save_buffer(cls, buffer: np.memmap[Any, np.dtype[Any]], path: str):
         # If buffer is already a memmap, we just need to flush it
         if buffer.filename != path:
             # If the paths are different, we need to create a new memmap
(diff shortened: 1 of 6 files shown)


sae_lens/training/activations_store.py
```diff
@@ -22,6 +21,30 @@
 from sae_lens.sae import SAE
 from sae_lens.tokenization_and_batching import concat_and_batch_sequences
 
+torch_to_numpy_dtype_dict: dict[torch.dtype, Any] = {
```
chanind, line 24: It looks like `bfloat16` is missing from this list. I don't think numpy supports bfloat16 though, even though it's a pretty common format for neural net training. Probably the safest would be to map it to `np.float32`, although it's sort of sad to need to use double the space for the numpy mmap in that case.

Lewington-pitsos: noice

jbloomAus:
Some notes:
- [ ] We've lost the 2x activation store getting shuffled in which may degrade results for SAE's trained with activations generated in real time. Maybe not worth retaining as it's messy code. 
- [ ] We need to pin the memory and move it to the SAE device asynchronously (eg: as [here](https://github.com/karpathy/nanoGPT/blob/9755682b981a45507f6eb9b11eadef8cb83cebd5/train.py#L116)). If we are loading straight from disk then the notion of the activation store having a device makes less sense and we should let the caller move it. 
- [ ] We should eliminate shuffling entirely in the cache activations runner and replace it with a specific shuffling script (per Anthropic's blog post). 
- [ ] Since we're clearly going for performance improvements, it might be nice to do ab test with the old code for a short run where we pre-cache activations and then train an SAE.

Lewington-pitsos:
Profiling has been completed on this branch: https://github.com/Lewington-pitsos/SAELens/tree/memmap_profiling in the `memmap_profiling` directory

it contains a `generate` script which creates activations and a `profile` script which loads 1000 batches of these activations. `torch profiler.profile` was used in each case on a cloud-based Nvidia A40 GPU.

The overall outcome is that the old script consumes around 26 CPU seconds and 3 GPU seconds, 
the new version with pinning consumes 3 CPU seconds and 11 GPU *milliseconds* 
and the new version without pinning consumes at 300 CPU milliseconds and 19 GPU milliseconds

This means that with no pinning we can achieve the following speeds:

```
Time taken to get 1000 batches: 0.39 seconds
Batches Per Second: 2538.15 batches per second
Samples Per Second: 649,766.1574397656 samples per second
Tokens Per Second: 665,360,545.21832 tokens per second
```

This is likely more than sufficient for training a 70B model. More thorough profiling might give us a clearer picture on the tradeoff between pinning and not pinning.

[memmap-pinning.txt](https://github.com/Lewington-pitsos/SAELens/blob/memmap_profiling/memmap_profiling/memmap-pinning.txt) - run with memmap and pinning as implemented on `4e8847b`. 
[memmap-no-pin.txt](https://github.com/Lewington-pitsos/SAELens/blob/memmap_profiling/memmap_profiling/memmap-no-pin.txt)- run with the condition above the pinning line always set to false, so instead of the pinning code we simply ran `tensor_batch = tensor_batch.to(self.device)`
`safetensors-9adba61b03b.txt](https://github.com/Lewington-pitsos/SAELens/blob/memmap_profiling/memmap_profiling/safetensors-9adba61b03b.txt) the same `generate` and `profile` ran from the main branch as of `9adba61b03b`

Note that the `profile` script appears to hang when testing the main branch on`9adba61b03b` for some reason pertaining to the torch profiling, but the profiling will complete after 20 minutes or so.

Lewington-pitsos:
I'm still not 100% sure what would be required to make this PR merge-able by the way

jbloomAus:
Huh, pinning slowed it down? I feel like somethings wrong here. Will read up and play with the code a bunch.

Lewington-pitsos:
Well, it led to 10x the CPU time but 1/2 the GPU time so possibly? I wasn't sure if a thorough comparison of pinning vs non-pinning was in scope for this PR

jbloomAus:
I think pinning maybe only helps when it's facilitating asynchronous work (ie: we're also doing SAE training) which you weren't doing here? So maybe expected? I'm inclined to say sorting this out is out of scope for this PR and probably I'll merge once I've checked the training an SAE with this kind of shuffling as opposed to our last isn't worse (for non-cached activation training).

Lewington-pitsos:
I actually had no idea what pinning even was before this PR so I feel underqualified to make a final judgement on it in the context of a system designed to do bleeding edge machine learning training XD

jbloomAus:
> I actually had no idea what pinning even was before this PR so I feel underqualified to make a final judgement on it in the context of a system designed to do bleeding edge machine learning training XD

I'm fairly happy with this but want to spend a bit of time on it. Sorry for the delay. Writing the shuffling utility will keep us moving if you've got time. Sorry for the delay here.

Lewington-pitsos:
no stress fam, I'm working on a bit of side-research at the moment to do with a possible metric for measuring SAE quality, I'll ping this board if I have time to work on shuffling. 

What I would basically do is make a shuffler which achieves approximately the same level of randomness as we had prior (but on disk)

Issue #222, opened 2024-07-09
[Proposal] Save norm scaling factor


ckkissane:
### Proposal 

After training, the scaling factor from ActivationsStore.estimate_norm_scaling_factor is lost. I propose storing it in wandb, or even just logging it as an MVP.

### Motivation

Lately i've trained multiple SAEs and want to quickly run evals on them. I can recompute the scaling factors myself, but this takes several minutes per each SAE.

### Checklist

- [x ] I have checked that there is no similar [issue](https://github.com/jbloomAus/SAELens/issues) in the repo (**required**)

jbloomAus:
solved.

PR #229, opened 2024-07-12
Trainer will now fold and log estimated norm scaling factor

curt-tigges:
# Description

Addresses https://github.com/jbloomAus/SAELens/issues/222. After performing a fit, the trainer will fold estimated_norm_scaling_factor into the SAE weights and will log this value to Weights and Biases.

## Type of change

# Checklist:


jbloomAus:
@curt-tigges I think it would be good to have code in the SAE Loader for pretrained saes such as Tom McGrath's GPT2 small residual stream SAEs where if we save the estimated scaling factor once to the yaml, then we don't need to re-estimate it before using the SAE (as we do [here](https://github.com/jbloomAus/SAELens/blob/db89dbc813e5a64a7609165474e3c66e2b537925/tests/benchmark/test_eval_all_loadable_saes.py#L89). 

To do:

- [ ] Remove change to wandb logging.
- [ ] Ensure we don't modify SAE during training to save with the factor folded in. 
- [ ] Add estimated cfg to yaml so an SAE saved with "expected_average_only_in" has it folded in and set to None when loading from pretrained.

curt-tigges:
@jbloomAus Rebased and made the changes you suggested. Value will not be logged in wandb. Scaling factor is folded in only after training, when saving the last checkpoint.

In addition, I did the following:
- Precomputed norm_scaling_factor for all SAEs with "normalize_activations" == "expected_average_only_in" and added it to the YAML
- Added a function to search the YAML for these values if needed
- Modified from_pretrained to identify if "normalize_activations" == "expected_average_only_in" in the config, and if so, executes the above function to find the norm_scaling_factor. It will then:
    - Fold norm_scaling_factor into the weights
    - Set cfg_dict["normalize_activations"] = "none"

jbloomAus:
Thanks!

PR #262, opened 2024-08-18
fix: fix memory crash / perf improvement when batching huge samples

chanind:
# Description

This PR fixes an issue where memory usage spikes and crashes the Python interpreter when batching extremely long samples. I found that I was unable to run pretokenization on the [monology/pile-uncopyrighted dataset](https://huggingface.co/datasets/monology/pile-uncopyrighted), and discovered that the reason is this dataset includes some extremely long text samples, upwards of 47 million characters in a single sample.

Our batching code previously would remove a single batch of tokens from the sample, and then return the batch and the remaining tokens. However, by returning the remaining tokens, this causes a lot of copies of this huge tensor, and seems to cause memory and performance issues.

To address this, this PR instead keeps track of an int offset within the sample and simply increments that offset as batches are formed, thus never directly modifying or copying the full sample. This means torch doesn't need to copy this potentially huge tensor over and over, saving memory and improving performance.

This also has the side-effect of speeding up batching and pretokenization since it avoids the unnecessary copying of the full sample tensor during batching.

## Type of change

# Checklist:

### DIFF
--- a/sae_lens/tokenization_and_batching.py
+++ b/sae_lens/tokenization_and_batching.py
@@ -6,20 +6,23 @@ import torch
 def _add_tokens_to_batch(
     batch: torch.Tensor | None,
     tokens: torch.Tensor,
+    offset: int,
     context_size: int,
     is_start_of_sequence: bool,
     begin_batch_token_id: int | None = None,
     begin_sequence_token_id: int | None = None,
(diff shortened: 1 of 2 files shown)


jbloomAus:
Looks great! thanks for catching and solving this!

PR #263, opened 2024-08-18
chore: adding more metatadata to pyproject.toml for PyPI

chanind:
# Description

This PR adds more metadata to pyproject.toml which PyPI will automatically parse and show on the [SAELens PyPI page](https://pypi.org/project/sae-lens/). This will add links to the git repo, homepage, keywords, topic, and license. This is minor, but doesn't hurt to add and makes the project look a bit more polished when people land on the PyPI.


jbloomAus:
thanks!

PR #265, opened 2024-08-23
fix: modify duplicate neuronpedia ids in config.yml, add test.

jbloomAus:
# Description

Some of the new neuronpedia id's weren't unique (mainly due to the SAEs from the instruction tuned model. Have fixed. 

## Type of change

### DIFF
--- a/sae_lens/pretrained_saes.yaml
+++ b/sae_lens/pretrained_saes.yaml
@@ -34,7 +34,7 @@ gpt2-small-res-jb:
     l0: 31.0
   - id: blocks.5.hook_resid_pre
     path: blocks.5.hook_resid_pre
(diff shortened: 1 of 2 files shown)


sae_lens/pretrained_saes.yaml
```diff
@@ -34,7 +34,7 @@ gpt2-small-res-jb:
     l0: 31.0
   - id: blocks.5.hook_resid_pre
     path: blocks.5.hook_resid_pre
-    neuronpedia: gpt2-small/0-res-jb
+    neuronpedia: gpt2-small/-res-jb
```
chanind, line 37: Is this intentional? It looks like this should be `5`?

PR #269, opened 2024-09-01
Tutorial: Corrected outdated code to call API

dhuynh95:
# Description

The code to call to the API of https://www.neuronpedia.org/api/explanation/export was outdated


jbloomAus:
thanks!

PR #273, opened 2024-09-02
minor dev improvement: return the original model as well inside of TrainingSAE

naterush:
# Description

Adds a new parameter to TrainingSAE - the model that this SAE was trained on. As SAETrainingRunner(cfg).run() may be called directly without previously defining the model, and TrainingSAE is return type, it can be useful to allow the user to access this original model directly from the returned SAE output.

Fixes # (issue)

## Type of change

- [x] New feature (non-breaking change which adds functionality)\

I do not think this is a breaking change, as TrainingSAE is only instantiated directly in one location in the codebase, and I do not think it is exposed internally to be created directly. But let me know if I'm missing something here. 

I added docs in the one place where they seemed relevant. Happy to move, though - let me know.

# Checklist:

(Again, check me on backwards compatibility please. Not totally sure about what the public interface is here. We could really easily fix backwards compatibility by making this parameter option, but we get less-strong types, so all things equal this solution seems better.). 

### Performance Check.

No changes here.

### DIFF
--- a/docs/training_saes.md
+++ b/docs/training_saes.md
@@ -211,3 +211,16 @@ upload_saes_to_huggingface(
     hf_repo_id="your-username/your-sae-repo",
 )
 ```
+
+## Accessing the model your SAE was trained on
+
+If you train a new SAE, and which to access the original model that your SAE was trained on directly, you can use the `original_model` property from the returned `sparse_autoencoder`:
+
+```python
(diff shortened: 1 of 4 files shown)


naterush:
Ok, this PR actually does break backwards compatibility through the TrainingSAEConfig.from_dict(config_dict) function. For that reason, I'm going to close this. Good experiment, I think!

PR #278, opened 2024-09-03
chore: reset hooks before saes in tutorial

chanind:
# Description

In the tutorial, the ablation code adds some hooks to an SAE via a `HookedSAETransformer`, but then detaches the SAE using `model.reset_saes()` before calling `model.reset_hooks()`. This causes the `model.reset_hooks()` call to not actually remove the ablation hook, thus permanently modifying the SAE for the rest of the notebook and breaking the attribution code which directly follows.

This PR swaps the order of these calls, so the `reset_hooks()` call now happens before the `reset_saes()` call.

## Type of change


jbloomAus:
Thanks! good catch!

PR #286, opened 2024-09-12
extend quick list creation options

TDulka:
# Description

Extends get_neuronpedia_quick_list with more options:
- passing a description and text to test across features
- passing a description of the sae to display instead of having to pass the whole SAE (sometimes you don't want to load the full SAE to just get a link to neuronpedia)
- giving more flexibility in what features you pick, should be able to make a list with features from different models and/or SAEs and provide a description for each feature

## Type of change

# Checklist:

### DIFF
--- a/sae_lens/analysis/neuronpedia_integration.py
+++ b/sae_lens/analysis/neuronpedia_integration.py
@@ -90,18 +90,21 @@ def get_neuronpedia_quick_list(
     default_test_text: Optional[str] = None,
 ) -> str:
     if isinstance(sae, SaeInfo):
+        if sae.neuronpedia_id is None:
+            raise ValueError(
+                "SAE does not have a Neuronpedia ID. Either dashboards for this SAE do not exist (yet) on Neuronpedia, or the SAE was not loaded via the from_pretrained method"
+            )
+
         neuronpedia_id = sae.neuronpedia_id
         model_name = sae.model_name
     elif isinstance(sae, SAE):
+        if sae.cfg.neuronpedia_id is None:
+            raise ValueError("Neuronpedia ID was not provided in SaeInfo.")
+
         neuronpedia_id = sae.cfg.neuronpedia_id
         model_name = sae.cfg.model_name
     else:
         raise TypeError("sae must be either SaeInfo or SAE")
-
-    if isinstance(sae, SAE) and neuronpedia_id is None:
-        raise ValueError(
-            "SAE does not have a Neuronpedia ID. Either dashboards for this SAE do not exist (yet) on Neuronpedia, or the SAE was not loaded via the from_pretrained method"
-        )
 
     if len(features) == 0:
         raise ValueError("No features provided")


TDulka:
make format wanted to change formatting of lots of old jupyter notebooks, I let it do it since it's in the checklist, but not sure what's your preferred approach

sae_lens/analysis/neuronpedia_integration.py
```diff
@@ -70,30 +70,71 @@ def open_neuronpedia_feature_dashboard(sae: SAE, index: int):
         webbrowser.open(url)
 
 
+class FeatureInfo(NamedTuple):
+    feature_index: int
+    description: str
+    model_name: Optional[str]
+    neuronpedia_id: Optional[str]
```
chanind, line 77: these should be default `None`, otherwise users will have to provide `model_name = None` explicitly

sae_lens/analysis/neuronpedia_integration.py
```diff
@@ -70,30 +70,71 @@ def open_neuronpedia_feature_dashboard(sae: SAE, index: int):
...
+        neuronpedia_id = sae.cfg.neuronpedia_id
+        model_name = sae.cfg.model_name
+    else:
+        raise TypeError("sae must be either SaeInfo or SAE")
 
-    sae_id = sae.cfg.neuronpedia_id
-    if sae_id is None:
-        print(
+    if isinstance(sae, SAE) and neuronpedia_id is None:
+        raise ValueError(
```
chanind, line 102: 💯

chanind (commented):
This PR is adding a lot of formatting changes to tutorial notebooks which don't seem to have any functionality change. Would it be possible to remove the notebooks from the PR that only have formatting changes?

tests/unit/analysis/test_neuronpedia_integration.py
```diff
@@ -69,3 +75,48 @@ async def test_neuronpedia_autointerp():
...
+    sae_info = SaeInfo(model_name="gpt2-small", neuronpedia_id="gpt2-small/0-res-jb")
+    features = [0, 1, 2]
+    url = get_neuronpedia_quick_list(sae_info, features, name="Test List")
+
+    expected_url = "https://neuronpedia.org/quick-list/?name=Test%20List&features=%5B%7B%22modelId%22%3A%20%22gpt2-small%22%2C%20%22layer%22%3A%20%220-res-jb%22%2C%20%22index%22%3A%20%220%22%7D%2C%20%7B%22modelId%22%3A%20%22gpt2-small%22%2C%20%22layer%22%3A%20%220-res-jb%22%2C%20%22index%22%3A%20%221%22%7D%2C%20%7B%22modelId%22%3A%20%22gpt2-small%22%2C%20%22layer%22%3A%20%220-res-jb%22%2C%20%22index%22%3A%20%222%22%7D%5D"
+    assert url == expected_url
+    mock_open.assert_called_once_with(expected_url)
+
+    # Reset mock
+    mock_open.reset_mock()
```
chanind, line 95: These are great tests! minor, but this is best split into two separate tests, one to test `get_neuronpedia_quick_list` with lists of features, one to test with detailed `FeatureInfo`. Names could be, e.g. `test_get_neuronpedia_quick_list_works_with_int_features` and `test_get_neuronpedia_quick_list_works_with_detailed_feature_info` or something similar

hijohnnylin:
thanks for writing this!
question: these changes are being drafted in preparation of an API update on Neuronpedia's side, correct? eg it's blocked on my side to implement the API support for this first.

sae_lens/analysis/neuronpedia_integration.py
TDulka: Thanks, fixed!

tests/unit/analysis/test_neuronpedia_integration.py
TDulka: Yeah, that makes sense, I split it into two.

TDulka:
@hijohnnylin Yes, this was indeed drafted in preparation for API update on Neuronpedia side, to give a sense how the query parameters can look like. Let me know if something doesn't make sense.

TDulka:
I removed the .ipynb formatting changes, they were indeed not appropriate for this PR.

jbloomAus:
Will attempt to make API support on our backend servers for this (search for a specific group of features) which I think will enable the api this will hit to be written / work efficiently :)

hijohnnylin:
Hey, while implementing the quick list with test text on the server side, I realized it's a bit difficult to cache these quick list activations, because the quick lists aren't stored in the database, nor are the activations that would be created. And we really do want to cache these activations, because we don't want to hit our GPUs every time that someone loads the page.

**Proposal: Use the (non-quick) Lists API to create these lists. The server-side already exists, though it was missing proper documentation. The documentation was created today.**

The difference is:
- POST call instead of GET call to create the list
  - the POST call will return a permanent link to the list that anyone can access
- You need a Neuronpedia API key, which you can get from "Settings" after logging in.

The documentation for doing this: https://www.neuronpedia.org/api-doc#tag/lists/POST/api/list/new-with-features
(the rest of the List docs are here: https://www.neuronpedia.org/api-doc#tag/lists)

Example code:
```
import requests

url = "https://www.neuronpedia.org/api/list/new-with-features"

payload = {
    "name": "Twitter features",
    "description": "This is a twitter list",
    "testText": "something about twitter accounts and Tweets. Follow them at @abc",
    "features": [
        {
            "modelId": "gpt2-small",
            "layer": "9-res-jb",
            "index": 1124
        },
        {
            "modelId": "gpt2-small",
            "layer": "2-att-kk",
            "index": 10116
        }
    ]
}
headers = {
    "Content-Type": "application/json",
    "X-Api-Key": "YOUR_API_KEY"
}

response = requests.post(url, json=payload, headers=headers)

print(response.json())
```

@TDulka @jbloomAus thoughts on this?

My apologies for not raising this earlier!

TDulka:
@hijohnnylin I'm not too familiar with the codebase, it seems from what I see that the advantage of these quick lists is that they don't rely on working with authentication and they are nice for just getting a quick view of a couple of features next to each other. I understand the point about the activations and caching making it complicated to include the test text in that. It seems in that case that it might not be appropriate to have the test text be a part of the quick list and if we'd like to have the test text it would be better to make another method which would incorporate authentication and use the api you mentioned.

My proposal would be to remove the test text from this PR to make it work without authentication and hitting the GPUs (if I understand it correctly). I think it could still be worth merging the rest to close this. We can leave the test text for another PR if someone still needs that feature.

PR #288, opened 2024-09-12
Improve error message for Gemma Scope non-canonical ID not found

ArthurConmy:
# Description

There are two issues with the current error message for SAE ID for not found:

1. The list of valid IDs can be absurdly long so doesn't even make the whole error fit:

Screenshot 2024-09-12 at 16 52 27

2. If users try to use e.g. `gemma-scope-2b-pt-res` we should give a hint that they might want to use `gemma-scope-2b-pt-res-canonical` instead so they don't need to specify `average_l0`. Here's a screenshot of the new behavior:

Screenshot 2024-09-12 at 17 05 56

~Fixes # (issue)~

## Type of change

# Checklist:

### Performance Check.

I have not implemented a training change.


jbloomAus:
thanks!

PR #292, opened 2024-09-13
Fix second issue with citation

ArthurConmy:
# Description

There was an extra `}` I missed. Also, ending all lines with commas is preferred (but not essential).

(Skipped rest of boilerplate)


chanind (approved):
Good catch!

PR #294, opened 2024-09-18
Support seqpos slicing

callummcdougall:
This allows seqpos slicing during training. Basically we add a `seqpos_slice` arg to the `LanguageModelSAERunnerConfig` (in the form of a tuple, which gets converted to a slice via `slice(*seqpos_slice)` - this is because slice objects aren't serializable when we're saving the config).

Apart from this config, the only other file getting changed is `activations_store.py`. It now has a `seqpos_slice` attribute, and it uses this to slice the activations which are fetched from `get_activations` (and which are used in `get_buffer`).

Note that the default behaviour is `seqpos_slice = (None,)`, which slices over all sequence positions. Also note that `seqpos_slice` can be used in conjunction with `context_size` (i.e. one doesn't make the other redundant).

### DIFF
--- a/sae_lens/config.py
+++ b/sae_lens/config.py
@@ -60,6 +60,7 @@ class LanguageModelSAERunnerConfig:
         store_batch_size_prompts (int): The batch size for storing activations. This controls how many prompts are in the batch of the language model when generating actiations.
         train_batch_size_tokens (int): The batch size for training. This controls the batch size of the SAE Training loop.
         normalize_activations (str): Activation Normalization Strategy. Either none, expected_average_only_in (estimate the average activation norm and divide activations by it -> this can be folded post training and set to None), or constant_norm_rescale (at runtime set activation norm to sqrt(d_in) and then scale up the SAE output).
+        seqpos_slice (tuple): Determines slicing of (batch, seq, d_in) activations when constructing batches, during training. Example: for Othello we sometimes use (5, -5).
         device (str): The device to use. Usually cuda.
         act_store_device (str): The device to use for the activation store. CPU is advised in order to save vram.
         seed (int): The seed to use.
@@ -151,6 +152,7 @@ class LanguageModelSAERunnerConfig:
     normalize_activations: str = (
         "none"  # none, expected_average_only_in (Anthropic April Update), constant_norm_rescale (Anthropic Feb Update)
     )
+    seqpos_slice: tuple[int | None, ...] = (None,)
 
     # Misc
     device: str = "cpu"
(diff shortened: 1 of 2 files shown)


jbloomAus:
Hey @callummcdougall  I've pushed:
- tests (in the future, please add tests!)
- ensured that this is in the SAE config (Since using an SAE / evaluating an SAE correctly will rely on knowing seqpos slice).
-  That we can serialize / deserialize without modification.

callummcdougall:
Got it, sorry for causing undue work - yes in the future will make sure to add tests! I wasn't sure about putting it in the sae config cause it's about the SAE's training data (or what inputs make sense for it) but not about e.g. the SAE's actual architecture. I was basing this on the fact that `ActivationsStore` gets initialized `from_config` which is a `LanguageModelSAERunnerConfig` not `SAEConfig` (although now I'm looking at that page, I see that it can also get initialized `from_sae`, so I get why this should be added).

jbloomAus:
@callummcdougall I think the idea is that if you couldn't evaluate the SAE without knowing about this property, then it needs to be in the SAE config. 

Speaking of which I don't see any changes to the evals.py but presumably we should ensure that evals are only run on seqpos positions? Are you able to do this?

tests/unit/training/test_activations_store.py
```diff
@@ -462,3 +462,26 @@ def test_validate_pretokenized_dataset_tokenizer_does_nothing_if_the_dataset_pat
...
+
+    activation_store = ActivationsStore.from_config(
+        ts_model, cfg, override_dataset=dataset
+    )
+
+    batch = activation_store.get_batch_tokens(1)
+    activations = activation_store.get_activations(batch)
+
+    assert batch.shape == (1, 10)  # Full context size
+    assert activations.shape == (1, 6, 1, cfg.d_in)  # Only 6 positions (2 to 7)
```
chanind, line 487: nice! Really great test 🥇

chanind (approved):
Code-wise this looks good to me, and looks like a reasonable addition to the library! Will defer to @jbloomAus if this is OK to merge

callummcdougall:
> Code-wise this looks good to me, and looks like a reasonable addition to the library! Will defer to @jbloomAus if this is OK to merge. I guess there's a question fo whether the expectation is that this would require different evals, or if this is something that only effects training.

Think it does seem valuable to also have the logged metrics during training only apply to the right sequence positions - is that what you meant @jbloomAus , or did you mean evals that are applied in a non-training context? Either way I can likely get to that later this week

PR #295, opened 2024-09-18
Fix gated forward functions

callummcdougall:
2 fixes:

1. In the `forward` method, when `self.use_error_term` is True, for non-standard architectures things like centering and input activation functions weren't being applied correctly. I think this was just a mistake, and is now fixed.
2. In the `encode_gated` method, the hook `hook_sae_acts_post` pointed to the activations post-ReLU (or whatever activation function is), but not to the actual output of this function, i.e. post-ReLU activations multiplied by masking values. I think a long-term solution has 2 separate hooks for each of these (e.g. `hook_sae_acts_post` and `hook_sae_mag_post`), but if we just have a single hook called `hook_sae_acts_post` then I think it makes a lot more sense for it to refer to the output of the encoder.


jbloomAus:
Thanks!

callummcdougall:
np! also let me know if there are things I can do to not require formatting PRs in the future - would this involve something like using the same workspace config files as this repo does?

jbloomAus:
@callummcdougall  run `make format` and `make check-ci` nothing more to it!

Issue #47, opened 2024-03-25
Don't allow setting both `d_sae` and `expansion_factor` in config


chanind:
These are both ways of setting the size of the SAE

anthonyduong9:
I can work on this this week.

anthonyduong9:
@chanind I was thinking that as part of this, we should also change

https://github.com/jbloomAus/SAELens/blob/a70822074c94c3bafe8502b54bc6e553483d08e3/sae_lens/config.py#L134

to

`    expansion_factor: Optional[int] = None`

since it's less explicit to check if `d_sae` is `4` than `None` to check if it's been set. Does that sound good to you?

jbloomAus:
seems good

chanind:
I think that sounds reasonable! But just make sure that if both `expansion_factor` is `None`, and `d_sae` is `None`, then `expansion_factor` should default to `4` for backwards compatibility. This can be noted in the docstring and also as a comment after `expansion_factor: Optional[int] = None  # defaults to 4 if d_sae and expansion_factor is None`

PR #301, opened 2024-09-24
Add value error if both d sae and expansion factor set

anthonyduong9:
# Description

Changes `LanguageModelSAERunnerConfig` so if `d_sae` and `expansion_factor` are not `None`, `__post_init__()` raises a `ValueError` and if `d_sae` and `expansion_factor` are `None`, `__post_init__()` sets `expansion_factor` to `4`. `d_sae` and `expansion_factor` are both attributes to set the size of an SAE, and `expansion_factor` has a default value of `4`, for backward compatibility.

Fixes # https://github.com/jbloomAus/SAELens/issues/47

## Type of change

# Checklist:

### Performance Check.

- [ ] L0
- [ ] CE Loss
- [ ] MSE Loss

### DIFF
--- a/tests/unit/helpers.py
+++ b/tests/unit/helpers.py
@@ -9,7 +9,7 @@ TINYSTORIES_MODEL = "tiny-stories-1M"
 TINYSTORIES_DATASET = "roneneldan/TinyStories"
 
 
-class ConfigKwargsType(TypedDict, total=False):
+class LanguageModelSAERunnerConfigDict(TypedDict, total=False):
     model_name: str
     hook_name: str
     hook_layer: int
@@ -45,7 +45,7 @@ def build_sae_cfg(**kwargs: Any) -> LanguageModelSAERunnerConfig:
     """
     Helper to create a mock instance of LanguageModelSAERunnerConfig.
     """
-    config_kwargs: ConfigKwargsType = {
+    mock_config_dict: LanguageModelSAERunnerConfigDict = {
         "model_name": TINYSTORIES_MODEL,
         "hook_name": "blocks.0.hook_mlp_out",
         "hook_layer": 0,
@@ -78,9 +78,9 @@ def build_sae_cfg(**kwargs: Any) -> LanguageModelSAERunnerConfig:
     }
 
     for key, value in kwargs.items():
-        config_kwargs[key] = value
+        mock_config_dict[key] = value
 
-    mock_config = LanguageModelSAERunnerConfig(**config_kwargs)
+    mock_config = LanguageModelSAERunnerConfig(**mock_config_dict)
 
     # reset checkpoint path (as we add an id to each each time)
     mock_config.checkpoint_path = (
--- a/tests/unit/helpers.py
+++ b/tests/unit/helpers.py
@@ -1,5 +1,5 @@
 import copy
-from typing import Any
+from typing import Any, Optional, TypedDict
 
 from transformer_lens import HookedTransformer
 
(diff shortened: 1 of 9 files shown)


tests/unit/helpers.py
```diff
@@ -9,51 +9,78 @@
...
-    for key, val in kwargs.items():
-        setattr(mock_config, key, val)
+    for key, value in kwargs.items():
+        mock_config_dict[key] = value
 
-    # Call the post-init method to set any derived attributes
-    # useful for checking the correctness of the configuration
-    # in the tests.
-    mock_config.__post_init__()
+    mock_config = LanguageModelSAERunnerConfig(**mock_config_dict)
```
chanind, line 83: fair enough! This feels less hacky than doing `__post_init__()` like before 👍

tests/unit/training/test_activations_store.py
```diff
@@ -28,7 +28,6 @@ def tokenize_with_bos(model: HookedTransformer, text: str) -> list[int]:
         {
             "model_name": "tiny-stories-1M",
             "dataset_path": "roneneldan/TinyStories",
-            "tokenized": False,
```
chanind, line 31: Why were these values changed?

tests/unit/training/test_sae_basic.py
```diff
@@ -202,7 +198,8 @@ def test_sae_save_and_load_from_pretrained_gated(tmp_path: Path) -> None:
 
 def test_sae_save_and_load_from_pretrained_topk(tmp_path: Path) -> None:
     cfg = build_sae_cfg(
-        activation_fn_str="topk", activation_fn_kwargs={"k": 30}, device="cpu"
```
chanind, line 205: why was this changed?

tests/unit/training/test_sae_trainer.py
```diff
@@ -204,7 +204,6 @@ def test_train_sae_group_on_language_model__runs(
     checkpoint_dir = tmp_path / "checkpoint"
     cfg = build_sae_cfg(
         checkpoint_path=str(checkpoint_dir),
-        train_batch_size=32,
```
chanind, line 207: Why was this changed?

chanind (commented):
The core code changes look good, but why were unrelated test values changed?

tests/unit/training/test_activations_store.py
anthonyduong9: Because `build_sae_cfg()` now takes a dict, updates another dict, and then passes it to a dataclass, whereas before it instantiated a dataclass, then looped over the dict and called `setattr()` to set the dataclass attributes. So now, since `"tokenized"` doesn't match any attributes of `LanguageModelSAERunnerConfig`, its `__init__()` method would return an error/tests would fail.

I checked throughout the codebase that there are no references to a field `tokenized` for any instance of `LanguageModelSAERunnerConfig`.

tests/unit/training/test_sae_basic.py
anthonyduong9: Similar reasons as in https://github.com/jbloomAus/SAELens/pull/301/files#r1776064470. `activation_fn_str` isn't an attribute of `LanguageModelSAERunnerConfig`.

tests/unit/training/test_sae_trainer.py
anthonyduong9: For similar reasons as in [this comment](https://github.com/jbloomAus/SAELens/pull/301/files#r1776064470). `train_batch_size` isn't an attribute of LanguageModelSAERunnerConfig.

tests/unit/helpers.py
anthonyduong9: Haha, yeah, we shouldn't effectively call `__post_init__()` twice.

tests/unit/training/test_activations_store.py
chanind: oooh nice, so if we mistype attributes in tests now we'll get an error? 🥇

chanind:
Awesome work with this, and thanks for fixing the issues with tests have invalid config fields set!

PR #304, opened 2024-09-25
fix: fixing canrager SAEs in SAEs table docs

chanind:
# Description

This PR fixes `docs/generate_sae_table.py` to support canrager SAEs, which is breaking when building docs. Currently, the code to load the config file for building the SAEs table re-implements the converter loading code, I think because the converters also download the full SAEs which is not necessary for building the table. 

This PR also removes `docs/sae_table.md` from git and adds it to `.gitignore` since this is auto-generated by CI anyway and the file in the repo is ignored. 

Longer-term, we should refactor the converters to allow loading only the config and not the full SAE so we don't have to remember to implement new converters in 2 places.

## Type of change

# Checklist:


jbloomAus:
thanks for the fix!

PR #309, opened 2024-10-01
Add entity argument to wandb.init

niniack:
# Description

The training runner does not respect the W&B entity argument, even when it is specified in the configuration.

This change just uses the already defined entity argument from the configuration and feeds it to `wandb.init()`.

## Type of change

# Checklist:


chanind (approved):
Good catch!

PR #310, opened 2024-10-01
Fix context size of a tokenized dataset

decandido:
# Description

Allowing for a `context_size` to be smaller than a dataset sample length in the case of a tokenized dataset. Adjusted tests to check for different cases of `context_size`. The motivation of this is to allow different applications like othelloGPT.

Please include a summary of the change and which issue is fixed. Please also include relevant motivation and context. List any dependencies that are required for this change.

Fixes # (issue)

## Type of change

# Checklist:

### DIFF
--- a/sae_lens/training/activations_store.py
+++ b/sae_lens/training/activations_store.py
@@ -219,6 +219,10 @@ class ActivationsStore:
                     f"""pretokenized dataset has context_size {ds_context_size}, but the provided context_size is {self.context_size}. 
                         The context_size {ds_context_size} is expected to be larger than or equal to the provided context size {self.context_size}."""
                 )
+            if self.context_size<0:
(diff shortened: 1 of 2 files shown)


chanind (approved):
This seems reasonable to me! Thanks for the test as well. Do you think it's worth warning that some data will be discarded if `ds_context_size != self.context_size`?

zhenningdavidliu:
Thats a good idea @chanind! We just noticed that in activation_store.py the package 'warnings' is not imported and the warnings are instead simply printed. Is there any particular reason for that? If not we are happy to rewrite the printed warnings as warnings.warn.

chanind:
No real reason we haven't used proper warnings - definitely want to switch things over to proper warnings / logs!

Issue #39, opened 2024-03-21
Support OthelloGPT


jbloomAus:
I did a hack weekend with Andy, Neel and others where we trained OthelloGPT SAEs fairly easily. 

I think it should be fairly easy to provide OthelloGPT support based on their code [here](https://github.com/jbloomAus/mats_sae_training/pull/8). The main trick will be to make an argument that tells the activation store not to do tokenization related stuff.

PR #317, opened 2024-10-03
Issue 39 support othellogpt in SAELens

decandido:
# Description

Adding support for OthelloGPT. Adding logic to train on sequences with a start and end position offset similar to [PR 294](https://github.com/jbloomAus/SAELens/pull/294). Adding tests for the added logic and to benchmark the OthelloGPT SAETrainingRunner.

Fixes # Issue [39](https://github.com/jbloomAus/SAELens/issues/39)

## Type of change

# Checklist:


jbloomAus:
Thanks @decandido and @zhenningdavidliu!

This has some overlap with https://github.com/jbloomAus/SAELens/pull/294. Specifically, I'd like to use the seqpos slicing instead of the start and end pos offsets. Other than that this looks good I think!

Requests:
- [] Please rebase or merge with https://github.com/jbloomAus/SAELens/pull/294 (which will also make sure the seqpos slice ends up in the config)
- [] Optional: Do you want to make a short tutorial on training / using the OthelloGPT SAE? 

Once that's done I'll accept :)

sae_lens/training/activations_store.py
```diff
@@ -451,27 +455,42 @@ def get_activations(self, batch_tokens: torch.Tensor):
 
         n_batches, n_context = batch_tokens.shape
 
-        stacked_activations = torch.zeros((n_batches, n_context, 1, self.d_in))
+        # For some models, we might want to exclude some positions from the sequence to train on
+        training_context_slice = list(
```
chanind, line 459: This is inefficient to create a list of indices like this. It's better to create an actual `slice` object, like `training_context_slice = slice(self.start_pos_offset, n_context - self.end_pos_offset)`. Alternative, you can probably just do something like the following directly when selecting the specific indices.

```python
stacked_activations[:, :, 0] = layerwise_activations[self.hook_name][
      :, start_pos:end_pos, self.hook_head_index
]
```

sae_lens/training/activations_store.py
```diff
@@ -491,10 +510,15 @@ def get_buffer(
         d_in = self.d_in
         total_size = batch_size * n_batches_in_buffer
         num_layers = 1
+        # Calculate the effective context size
+        training_context_slice = list(
+            range(self.start_pos_offset, context_size - self.end_pos_offset)
+        )
+        training_context_size = len(training_context_slice)
```
chanind, line 517: Here as well, this is inefficient to create a list from a range and then measure it. This would be better to just directly compute like `training_context_size = context_size - self.end_pos_offset - self.start_pos_offset`. This could also be moved to a shared function since it looks like it's shared with code in `get_activations()`

tests/unit/training/test_cache_activations_runner.py
```diff
@@ -216,3 +216,89 @@ def W_E(self) -> torch.Tensor:
...
+    n_batches_in_buffer = 32
+    print(f"n batches in buffer: {n_batches_in_buffer}")
+    store_batch_size = 1
+    print(f"store_batch_size: {store_batch_size}")
+    n_buffers = 3
+    print(f"n_buffers: {n_buffers}")
+
+    tokens_in_buffer = n_batches_in_buffer * store_batch_size * training_context_size
+    total_training_tokens = n_buffers * tokens_in_buffer
+    print(f"Total Training Tokens: {total_training_tokens}")
```
chanind, line 246: these print statements should be removed for tests

sae_lens/training/activations_store.py
```diff
@@ -451,27 +455,42 @@ def get_activations(self, batch_tokens: torch.Tensor):
...
+                :, training_context_slice, self.hook_head_index
             ]
         elif (
             layerwise_activations[self.hook_name].ndim > 3
         ):  # if we have a head dimension
             try:
                 stacked_activations[:, :, 0] = layerwise_activations[
                     self.hook_name
-                ].view(n_batches, n_context, -1)
+                ].view(n_batches, training_context_size, -1)[
```
chanind, line 479: This looks like it's not correct - don't we want to first stack into `n_context` sequences and then lop off the start and end? It looks like as this is written it will turn this into `training_context_size` sequences and then further slice them, likely leading to the wrong size sequences. Is this covered by a test?

chanind (commented):
If I userstand this correctly, these offset parameters only makes sense in cases where the model is given sequences of a fixed length, and `context_size` is set to that length, e.g. OthelloGPT?

sae_lens/training/activations_store.py
decandido: Hey David, thanks for your thorough review of our PR! As Joseph mentioned in his [comment](https://github.com/jbloomAus/SAELens/pull/317#issuecomment-2391999612) above, we've rebased on Callum's PR and added a few tests of our own. These problems are not present in the rebased code, but they are really helpful for @zhenningdavidliu and me! 🙏

decandido:
> Requests:
> 
> * [] Please rebase or merge with [Support seqpos slicing #294](https://github.com/jbloomAus/SAELens/pull/294) (which will also make sure the seqpos slice ends up in the config)
> * [] Optional: Do you want to make a short tutorial on training / using the OthelloGPT SAE?

Thanks @jbloomAus for reviewing our PR! We've rebased on #294 and removed the start and end pos offsets. 

We also added a short script to train SAEs on othelloGPT (`scripts/training_a_sparse_autoencoder_othelloGPT.py`). Let us know if that's what you were looking for!

callummcdougall:
Hey, thanks for all your awesome work here! Could I possibly bump on this PR? Just cause it would also be super useful for the final bits of the ARENA material to run smoothly (-:

callummcdougall:
Thanks! (I'm actually also realizing that the context length issue wasn't a part of this PR, it was an earlier one, and that was the main thing that was making things a bit clunky, so less important thank I thought - but would still be great!)

PR #331, opened 2024-10-13
Dedupe code to get config

### LINKED ISSUE #305: [Proposal] Refactor `generate_sae_table.py` and `pretrained_sae_loaders.py` to share code
### Proposal 

Currently, the converters in `pretrained_sae_loaders.py` cannot be used in `generate_sae_table.py` because they return both a config file and the full SAE. As a result, we need to remember to manually re-create a lot of the functionality of the converters in `generate_sae_table.py` every time we add a new converter, and it's easy to forget / have these code paths be subtly different.

We should refactor the converter interface so there's a way to return only the config so that `generate_sae_table.py` can use the same converter code path as in `pretrained_sae_loaders`, and we don't need to manually duplicate code between these two files.

### Checklist

### LINKED ISSUE #315: [Proposal] Simple get_sae_config method for pretrained SAEs
### Proposal 

I would like a simple `get_sae_config` that takes the `release` and `id` (under `saes`), and returns the config for the SAE.
This should not require downloading the whole SAE weights.

### Motivation

Use case:
If someone is looking to get the config to answer a quick question (eg what was the width of that one SAE?), they shouldn't need to have any knowledge of these conversion functions, or be forced to load the weights into memory.

Currently, `PretrainedSaeLoader` loads four different "types" of SAEs (each with a different conversion function): default (no conversion func), connor_rob_hook_z, gemma_2, and dictionary_learning_1.

To get just the config for each of them, I have to special case each of them, like so:
```
        if conversion_func is None:
            config = get_sae_config_from_hf(
                item.repo_id, folder_name, force_download=FORCE_DOWNLOAD
            )
        elif conversion_func == "connor_rob_hook_z":
            print("connor_rob_hook_z")
            # we don't have a special function to just get the config, so we have to download the whole thing
            config, _, _ = connor_rob_hook_z_loader(
                item.repo_id, folder_name, force_download=FORCE_DOWNLOAD
            )
        elif conversion_func == "dictionary_learning_1":
            print("dictionary_learning_1")
            # we don't have a special function to just get the config, so we have to download the whole thing
            config, _, _ = dictionary_learning_sae_loader_1(
                item.repo_id, folder_name, force_download=FORCE_DOWNLOAD
            )
        elif conversion_func == "gemma_2":
            print("gemma_2")
            config = get_gemma_2_config(item.repo_id, folder_name)
```

There are some problems with this:
1) Difficult to maintain - Requires updating my function every time there is a new conversion function.
2) Unnecessary loading weights - The `connor_rob_hook_z` and `dictionary_learning_1` conversion types require downloading (and sometimes loading) the weights of the SAEs. I don't want to do this when all I want is the config.

### Pitch

A simple `get_sae_config` method that takes two arguments: `release_id` and `sae_id` (and optionally, force_download if we suspect the config has changed). It returns the config without downloading weights or loading them.

### Alternatives

The current alternative is to write the special cases as above.

### Checklist

anthonyduong9:
# Description

Adds a new function `get_sae_config()` which takes a `release` and `sae_id` and returns the config for an SAE. To get the config for an SAE, one shouldn't have to know the conversion function or download the weights, and one is more likely to know the `release` and `sae_id` than `repo_id` and `folder_name`. We also currently duplicate a lot of code across `pretrained_sae_loaders.py` and `generate_sae_table.py`.

Fixes #305, fixes #315 

## Type of change

# Checklist:

### Performance Check.

- [ ] L0
- [ ] CE Loss
- [ ] MSE Loss

### DIFF
--- a/sae_lens/toolkit/pretrained_sae_loaders.py
+++ b/sae_lens/toolkit/pretrained_sae_loaders.py
@@ -130,78 +130,15 @@ def handle_config_defaulting(cfg_dict: dict[str, Any]) -> dict[str, Any]:
     return cfg_dict
 
 
-def connor_rob_hook_z_loader(
-    repo_id: str,
-    folder_name: str,
-    device: Optional[str] = None,
-    force_download: bool = False,
-    cfg_overrides: Optional[dict[str, Any]] = None,
-) -> tuple[dict[str, Any], dict[str, torch.Tensor], None]:
-
-    file_path = hf_hub_download(
-        repo_id=repo_id, filename=folder_name, force_download=force_download
-    )
+def get_connor_rob_hook_z_config(
+    repo_id: str, folder_name: str, device: Optional[str]
+) -> dict[str, Any]:
     config_path = folder_name.split(".pt")[0] + "_cfg.json"
     config_path = hf_hub_download(repo_id, config_path)
+
     old_cfg_dict = json.load(open(config_path, "r"))
 
-    weights = torch.load(file_path, map_location=device)
-    # weights_filename = f"{folder_name}/sae_weights.safetensors"
-    # sae_path = hf_hub_download(
-    #     repo_id=repo_id, filename=weights_filename, force_download=force_download
-    # )
-    # if device is None:
-    #     device = "cuda" if torch.cuda.is_available() else "cpu"
-
-    # return load_pretrained_sae_lens_sae_components(cfg_path, sae_path, device)
-
-    # old_cfg_dict = {
-    #     "seed": 49,
-    #     "batch_size": 4096,
-    #     "buffer_mult": 384,
-    #     "lr": 0.0012,
-    #     "num_tokens": 2000000000,
-    #     "l1_coeff": 1.8,
-    #     "beta1": 0.9,
-    #     "beta2": 0.99,
-    #     "dict_mult": 32,
-    #     "seq_len": 128,
-    #     "enc_dtype": "fp32",
-    #     "model_name": "gpt2-small",
-    #     "site": "z",
-    #     "layer": 0,
-    #     "device": "cuda",
-    #     "reinit": "reinit",
-    #     "head": "cat",
-    #     "concat_heads": True,
-    #     "resample_scheme": "anthropic",
-    #     "anthropic_neuron_resample_scale": 0.2,
-    #     "dead_direction_cutoff": 1e-06,
-    #     "re_init_every": 25000,
-    #     "anthropic_resample_last": 12500,
-    #     "resample_factor": 0.01,
-    #     "num_resamples": 4,
-    #     "wandb_project_name": "gpt2-L0-20240117",
-    #     "wandb_entity": "ckkissane",
-    #     "save_state_dict_every": 50000,
-    #     "b_dec_init": "zeros",
-    #     "sched_type": "cosine_warmup",
-    #     "sched_epochs": 1000,
-    #     "sched_lr_factor": 0.1,
-    #     "sched_warmup_epochs": 1000,
-    #     "sched_finish": True,
-    #     "anthropic_resample_batches": 100,
-    #     "eval_every": 1000,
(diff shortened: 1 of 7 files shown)


sae_lens/toolkit/pretrained_saes_directory.py
```diff
@@ -74,3 +74,11 @@ def get_norm_scaling_factor(release: str, sae_id: str) -> Optional[float]:
                 if sae_info["id"] == sae_id:
                     return sae_info.get("norm_scaling_factor")
     return None
+
+
+def get_repo_id_and_folder_name(release: str, sae_id: str) -> tuple[str, str]:
```
chanind, line 79: nit: should add a unit test for this

sae_lens/toolkit/pretrained_sae_loaders.py
```diff
@@ -451,32 +449,50 @@ def get_dictionary_learning_config_1(config: dict[str, Any]) -> dict[str, Any]:
...
+    if conversion_loader_name not in NAMED_PRETRAINED_SAE_LOADERS:
+        raise ValueError(
+            f"Conversion func '{conversion_loader_name}' not found in NAMED_PRETRAINED_SAE_LOADERS."
+        )
+    return conversion_loader_name
+
+
+def get_sae_config(
+    release: str, sae_id: str, options: SAEConfigLoadOptions
+) -> dict[str, Any]:
```
chanind, line 465: nit: this function should be covered by unit tests. benchmark tests don't get run in CI so don't really add any guarantees to the codebase. It should be cheap to load configs in a unit test since we don't need to download the whole model weights now. This could look like trying to load a specific config for each of the types of SAEs we have loaders for, and just asserting that the resulting config dict matches what we expect.

chanind (commented):
Overall this looks great! Just could use some unit tests on the new get config function and should be good to go!

sae_lens/toolkit/pretrained_sae_loaders.py
anthonyduong9: I've added tests for each item in `NAMED_PRETRAINED_SAE_LOADERS`. Have we thought about splitting the tests in the `unit` directory into unit tests and integration tests?

The tests I've added are technically integration tests since they're interacting with the filesystem and downloading from Hugging Face. Though I think it's more important for the tests to be useful than to fit the definition of "unit tests".

sae_lens/toolkit/pretrained_saes_directory.py
anthonyduong9: I've added tests.

anthonyduong9:
> Overall this looks great! Just could use some unit tests on the new get config function and should be good to go!

Thanks!

sae_lens/toolkit/pretrained_sae_loaders.py
chanind: Definitely agree the word "unit" is not technically correct and is probably confusing, but I also don't think proper unit vs integration split matters here as the codebase isn't that large and we don't want to, e.g., only run the unit and not integration tests to save time. The majority of the tests in "test/unit" are technically integration tests since we don't mock out dependencies unless we absolutely have to and will happily run against real (tiny) models / data in tests.

IMO we just need to make sure functionality is covered by tests, whether unit or integration or both, and we shouldn't worry much about the distinction. I would normally just say "tests"  instead of "unit tests", but since we have a folder "tests/unit" which is run by CI and "tests/benchmark" which isn't, want to emphasize the "test/unit" folder. I'm not sure what to call the non-benchmark tests that isn't confusing, but also want as much as possible for people to add to tests that are run by CI and not add tests to "test/benchmark" which are never run.

Maybe we can move "tests/benchmark" into just "benchmark", and then move "tests/unit" into "tests", and then delete the "tests/acceptance" folder which is empty anyway? Probably worth discussing in an issue.

hijohnnylin:
@anthonyduong9 just wanted to say i used this today and it worked great! thank you.

PR #328, opened 2024-10-10
fix: Test JumpReLU/Gated SAE and fix sae forward with error term

chanind:
# Description

This PR adds test coverage to the JumpReLU and GatedSAE encode methods.

In doing this, I realized there's a lot of duplication  between all the encode variants and cleaned that up as well. I think there were some potential minor bugs in this duplication, for instance in `forward()`, when adding error term, we called `run_time_activation_norm_fn_out()`  after `reshape_fn_out()`, but we do the opposite in `decode()`.

It looks like most of the duplication between the `error` section of `forward()` and the normal `encode() / decode()` was there just to avoid triggering hooks, so I added a contextmanager `_disable_hooks()` which is used to disable hooks in this branch of the code while reusing our existing `encode() / decode()` methods. This should mean we don't need to worry about these duplicated codepaths diverging and causing bugs. 

If the refactor is out of scope, I can revert the changes to `sae.py` and just leave the test coverage.

Fixes #323 
Fixes #326 

Note: after investigating #326, it looks like this is caused by a bad copy/paste in the duplicated code for jumprelu forward, where we're accidentally including the hooks: https://github.com/jbloomAus/SAELens/blob/main/sae_lens/sae.py#L479.

## Type of change

Please delete options that are not relevant.

- [x] refactor
- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

# Checklist:

- [x] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] My changes generate no new warnings
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes
- [x] I have not rewritten tests relating to key interfaces which would affect backward compatibility

### You have tested formatting, typing and unit tests (acceptance tests not currently in use)

- [x] I have run `make check-ci` to check format and linting. (you can run `make format` to format code if needed.)

PR #334, opened 2024-10-16
Encode with slice

callummcdougall:
It's useful to be able to get a slice of latent activations in an architecture-general way, without having to compute all the latent activation values (e.g. I've found this useful while working on autointerp for SAEBench, which only samples a fraction of the total latents). I've implemented this by adding an optional `latents` argument to all the `encode` functions - e.g. if we supply `latents=range(100)` then we'll just compute feature activations for the first 100 features. The default is `latents=None` which works as normal.

Some other notes:

- I've added a test function to `test_sae_basic.py`, which just verifies that "run `encode` with `latents`" gives the same result as "run `encode` then slice the result with `latents`"
- I'm aware this won't work for TopK, but we can add an error message when it comes to that

Also a somewhat unrelated thing - I'm not sure why the `forward` function doesn't use `encode` (we essentially have the `encode` code duplicated here) - if there's no good reason for this, then I'm happy to submit a PR to fix this.

### DIFF
--- a/sae_lens/training/training_sae.py
+++ b/sae_lens/training/training_sae.py
@@ -5,7 +5,7 @@ https://github.com/ArthurConmy/sae/blob/main/sae/model.py
 import json
 import os
 from dataclasses import dataclass, fields
-from typing import Any, Iterable, Optional
+from typing import Any, Optional
 
 import einops
 import torch
@@ -200,10 +200,11 @@ class TrainingSAE(SAE):
             assert self.use_error_term is False, "Gated SAEs do not support error terms"
 
     def encode_standard(
-        self, x: Float[torch.Tensor, "... d_in"], latents: Iterable[int] | None = None
+        self, x: Float[torch.Tensor, "... d_in"], latents: torch.Tensor | None = None
     ) -> Float[torch.Tensor, "... d_sae"]:
         """
-        Calcuate SAE features from inputs
+        Calcuate SAE features from inputs. The `latents` argument is ignored (this is just so the type signature matches
+        the parent class, which uses this argument to compute only a subset of the SAE hidden values)
         """
         assert (
             latents is None


sae_lens/sae.py
```diff
@@ -544,11 +560,12 @@ def encode_jumprelu(
         return feature_acts
 
     def encode_standard(
-        self, x: Float[torch.Tensor, "... d_in"]
+        self, x: Float[torch.Tensor, "... d_in"], latents: Iterable[int] | None = None
```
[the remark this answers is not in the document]
chanind: It also looks like `topk` SAEs go through this codepath and will silently break if anything is passed for `latents`. We should make it non-silent.

chanind (commented):
How does this interact with error term calculation, or is this only a performance optimization when not running a full forward pass? How significant of a performance speed-up in your use-case?

I can see the benefit of this, but also worry it makes encode() more complicated if users decide to make SAE subclasses, adds a confusing feature to the `TrainingSAE.encode()` where it looks like you can pass a `latents` arg but then it throws an error if you try to, and silently breaks `topk` SAEs. If the performance improvement this gives is worth the added complexity / downsides it can still make sense to do though.

chanind:
Re: the separate encode implementations in forward, I noticed this as well and fixed it in https://github.com/jbloomAus/SAELens/pull/328, just waiting on a review. It seems like the reason is that we didn't want to trigger hooks in error term calculation since then the error term would change when users do interventions on SAE latents. But it also seems like all this duplication is likely introducing subtle bugs

callummcdougall:
> How does this interact with error term calculation, or is this only a performance optimization when not running a full forward pass? How significant of a performance speed-up in your use-case?

It's designed only for `encode`, I don't anticipate a big use case for people doing full fwd passes. It's specifically if you want a way of computing a small subset of feature activations without computing all of them, because otherwise if we have a tensor of `(batch, seq_len, d_sae)` with very large `d_sae`, this can dominate memory costs and cause OOMs (which is obvs impractical if we only want e.g. 1 feature and `d_sae` is O(10k) or larger).

> nit: should this also take a tensor as well? It seems like the code will work with a tensor of ints as well

Yep, I can make that change! Makes sense that this would be a bit easier for the type errors. I'll do that in my next commit.

> It also looks like topk SAEs go through this codepath and will silently break if anything is passed for latents. We should make it non-silent.

Ah thanks, didn't realize TopK already existed & routed through that, yep I'll get that sorted out as well.

> nit: It would be good to test with pytest.mark.parametrize with all architectures, since it seems like it's easy to accidentally mess up the implementation in one architecture and have it slip through the cracks

Yep will sort that out

callummcdougall:
> I can see the benefit of this, but also worry it makes encode() more complicated if users decide to make SAE subclasses, adds a confusing feature to the TrainingSAE.encode() where it looks like you can pass a latents arg but then it throws an error if you try to, and silently breaks topk SAEs

I can add a short docstring explaining about this. Joseph mentioned he implemented this once using a wrapper around the other encode functions but I think it really has to be done on a per-architecture basis, because things like TopK (and possibly more complicated architectures) need to be considered as special cases.

callummcdougall:
@chanind most recent test has failed because `LanguageModelSAERunnerConfig` has a different name for the activation function (it calls it `activation_fn` and doesn't have an `activation_fn_str`, whereas `SAEConfig` has an `activation_fn_str` and `activation_fn` is the actual activation). Not pushing a fix yet because I don't have full context on why these are different and I don't think I'll be able to guess what the best solution is without that context. Do you have a sense for what the best solution here is?

sae_lens/sae.py
```diff
@@ -491,9 +488,39 @@ def forward(
...
+        """
+        Computes the latent values of the Sparse Autoencoder (SAE) using a gated architecture. The activation values are
+        computed as the product of the masking term & the post-activation function magnitude term:
+
+            1[(x - b_dec) @ W_gate + b_gate > 0] * activation_fn((x - b_dec) @ W_enc + b_enc)
+
+        The `latents` argument allows for the computation of a specific subset of the hidden values. If `latents` is not
+        provided, all latent values will be computed.
+        """
+        latents_tensor = torch.arange(self.cfg.d_sae) if latents is None else latents
```
chanind, line 523: The previous implementation you had, `latents_slice = slice(None) if latents is None else torch.tensor(latents)`, seems better to me since this new version will create a new tensor of size `d_sae` on every SAE forward pass when not selecting specific latents. This would likely reduce performance for most users, which seems counter-productive since this PR is just meant to be a performance improvement if I understand the goal correctly. Wouldn't the old implementation have worked fine just adding `torch.Tensor` to the type of `latents`? e.g. `latents: Iterable[int] | torch.Tensor | None = None`

sae_lens/sae.py
```diff
@@ -164,17 +154,8 @@ def __init__(
...
-            self.encode = self.encode_standard
-        elif self.cfg.architecture == "gated":
-            self.initialize_weights_gated()
-            self.encode = self.encode_gated
-        elif self.cfg.architecture == "jumprelu":
-            self.initialize_weights_jumprelu()
-            self.encode = self.encode_jumprelu
-        else:
-            raise (ValueError)
+        if self.cfg.architecture not in ["standard", "gated", "jumprelu"]:
```
chanind, line 157: It looks like tests are failing because `initialize_weights()` have been moved into their own function, but that function is never called now. IMO this error can be raised in the new `initialize_weights()` method too.

chanind (commented):
I don't follow the problem related to `activation_fn` for the runner config vs `activation_fn_str` for the SAE config. I agree it's weird these are different and don't know why they're not the same, but why is it causing a problem? It looks like tests are failing currently because `initialize_weights()` is not being called, it doesn't look related to `activation_fn_str`.

jbloomAus:
Sharing a link to my previous solution for this: https://github.com/jbloomAus/SAEDashboard/blob/4f3d6e0d0320816f10de7045dc8e3cbc3ee29f0e/sae_dashboard/feature_data_generator.py#L220

@callummcdougall @chanind can I leave this with you both? Sorry for being MIA. Think this is a good idea.

callummcdougall:
Ah really sorry forgot to follow up on this - yep I'll try and get this sorted this week (and apologies if I'd got wrong the reason for failing tests)

adamkarvonen:
I think it would be better to do Joseph's solution (just modifying W_enc, threshold, etc) rather than encoding a subset. It's simpler, and I believe non contiguous indexing into the encoder is actually pretty expensive in terms of memory bandwidth. If it wasn't, it would be faster to index into W_dec rather than doing the full sparse matmul. I think `encode_subset()` would probably be slower than `encode()`, although it would use less memory.

Issue #320, opened 2024-10-06
Use Huggingface Datasets for ActivationStore


tom-pollak:
### Proposal 

Use huggingface Dataset rather than safetensors caching activations ActivationStore.

### Motivation

Huggingface datasets has many of the features included that we'd be interested in including:

- Consistent format, easily to load and share on hf hub.
- Caching from multiple hook points at once, with different shapes, in a single dataset (as different columns)
- Simplified: Sharding and memory mapping is default in `datasets.load_from_disk
- Datasets can be loaded as IterableDataset / streaming=True / n_shards, much simpler than doing manually.
- Shuffling support would be [easier](https://huggingface.co/docs/datasets/v3.0.1/en/package_reference/main_classes#datasets.IterableDataset.shuffle)

#### Schema outline

I intend to create a dataset with a schema:

```typescript
{
  "hook_point_1": Array2D(shape=(T, D_1)),
  "hook_point_2": Array2D(shape=(T, D_2)) # notably different shape
}
```

(currently we only allow a single hook point when collecting cached activations, mainly because it gets a lot more complicated doing it manually)

I'll create a [sharded dataset](https://github.com/huggingface/datasets/issues/4091#issuecomment-1089101841) using huggingface.

### Alternatives

Keep current implementation, implement activation dataset manually manually.

### Checklist

- [X] I have checked that there is no similar [issue](https://github.com/jbloomAus/SAELens/issues) in the repo (**required**)

PR #336, opened 2024-10-17
chore: reduce test space usage in CI

chanind:
# Description

CI has started failing since merge #320 due to running out of space. It looks like this is due to loading and processing large datasets (c4-tokenized-2b). This PR replaces that dataset with a tiny tokenized version of c4-10k: https://huggingface.co/datasets/chanind/c4-10k-mini-tokenized-16-ctx-gelu-1l-tests. This is a tokenized version of the first 1k rows of the c4-10k dataset. It's split into 64 pieces, and the total dataset size is onlly 250kb (vs 2gb for c4-tokenized-2b)

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -32,16 +32,16 @@ jobs:
       - name: Cache Huggingface assets
         uses: actions/cache@v4
         with:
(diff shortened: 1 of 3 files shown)


chanind:
merging as this should be uncontroversial and CI is currently failing due to space issues.

PR #340, opened 2024-10-19
changes dtype default value in read_sae_from_disk()

### LINKED ISSUE #230: [Bug Report] read_sae_from_disk should default to the SAE dtype unless it's overridden
**Describe the bug**
`sae_lens_loader` in `pretrained_sae_loaders.py` currently does not pass in `dtype` into `read_sae_from_disk`, so it always initializes weights using `float32` first.

The dtype is correct at the very end, since `sae` is created/initialized with the correct dtype from cfg_dict. However, ideally we would be using the correct dtype the whole time.

The fix seems to be that `read_sae_from_disk` should 1) make dtype optional and 2) use the dtype specified by the cfg_dict by default, if dtype is not specified. This has the benefit that other methods calling `read_sae_from_disk` would have the correct behavior by default, unless they choose to override.

I'll make a PR for this soon.

### Checklist

anthonyduong9:
# Description

Changes `read_sae_from_disk()` so that the `dtype` param is of type `Optional[torch.dtype]` and has a default value of `None`, in which case the function basically sets `dtype` to `cfg_dict["dtype"]`. `dtype` should be `cfg_dict["dtype"]`, unless the caller specifies a different value.

Fixes #230 

## Type of change

# Checklist:

### Performance Check.

- [ ] L0
- [ ] CE Loss
- [ ] MSE Loss

### DIFF
--- a/sae_lens/sae.py
+++ b/sae_lens/sae.py
@@ -642,7 +642,6 @@ class SAE(HookedRootModule):
             cfg_dict=cfg_dict,
             weight_path=weight_path,
             device=device,
-            dtype=DTYPE_MAP[cfg_dict["dtype"]],
         )
 
         sae_cfg = SAEConfig.from_dict(cfg_dict)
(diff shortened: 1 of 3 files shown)


chanind (approved):
👍 LGTM

PR #347, opened 2024-10-23
Supply `device` to `SAEConfigLoadOptions`

callummcdougall:
This is a bug which was introduced last week. The issue seems to be that the `SAEConfigLoadOptions` class was introduced which contains the `device` field (default `None`), and this field wasn't set when device is not None. Before this change, the config was just loaded directly from HuggingFace without overriding its `device` value (which on HuggingFace presumably is "cuda").

Would be great if this PR could be merged soon if possible, and I can try to add tests for this kind of failure mode afterwards.


chanind:
Thanks for finding this! just merged as this is a live bug. If you can add a test that would prevent this regression in the future that would be great as well

callummcdougall:
Yep will add a test on my todo list for this week!

PR #352, opened 2024-10-30
Add logic to train JumpReLU SAEs

### LINKED ISSUE #330: [Proposal] Support training JumpReLU SAEs
### Proposal 

We support running pre-trained JumpReLU SAEs, but we don't currently support them in training (afaik). These are a SOTA architecture so would be great if we could support this in training.

### Checklist

anthonyduong9:
# Description

Adds logic to train JumpReLU SAEs. JumpReLU is a state-of-the-art architecture, so users should be able to train JumpReLU SAEs. Currently, they can load pre-trained JumpReLU SAEs and perform inference with them.

Fixes #330

## Type of change

# Checklist:

### Performance Check.

- [ ] L0
- [ ] CE Loss
- [ ] MSE Loss

### DIFF
--- a/sae_lens/config.py
+++ b/sae_lens/config.py
@@ -162,7 +162,7 @@ class LanguageModelSAERunnerConfig:
     seed: int = 42
     dtype: str = "float32"  # type: ignore #
     prepend_bos: bool = True
-    initial_threshold: float = 0.5
+    initial_threshold: float = 0.001
 
     # Performance - see compilation section of lm_runner.py for info
     autocast: bool = False  # autocast to autocast_dtype during training
@@ -411,6 +411,7 @@ class LanguageModelSAERunnerConfig:
             "decoder_heuristic_init": self.decoder_heuristic_init,
             "init_encoder_as_decoder_transpose": self.init_encoder_as_decoder_transpose,
             "normalize_activations": self.normalize_activations,
+            "initial_threshold": self.initial_threshold,
         }
 
     def to_dict(self) -> dict[str, Any]:
--- a/sae_lens/config.py
+++ b/sae_lens/config.py
@@ -66,7 +66,7 @@ class LanguageModelSAERunnerConfig:
         seed (int): The seed to use.
         dtype (str): The data type to use.
         prepend_bos (bool): Whether to prepend the beginning of sequence token. You should use whatever the model was trained with.
-        threshold (float): Threshold for training JumpReLU SAEs.
+        jumprelu_init_threshold (float): The threshold to initialize for training JumpReLU SAEs.
         autocast (bool): Whether to use autocast during training. Saves vram.
         autocast_lm (bool): Whether to use autocast during activation fetching.
         compile_llm (bool): Whether to compile the LLM.
@@ -163,7 +163,7 @@ class LanguageModelSAERunnerConfig:
     seed: int = 42
     dtype: str = "float32"  # type: ignore #
     prepend_bos: bool = True
-    threshold: float = 0.001
+    jumprelu_init_threshold: float = 0.001
 
     # Performance - see compilation section of lm_runner.py for info
     autocast: bool = False  # autocast to autocast_dtype during training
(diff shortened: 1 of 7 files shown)


sae_lens/training/training_sae.py
```diff
@@ -24,6 +26,68 @@
...
+        x, threshold, bandwidth = inputs
+        del output
+        ctx.save_for_backward(x, threshold)
+        ctx.bandwidth = bandwidth
+
+    @staticmethod
+    def backward(ctx: Any, grad_output: torch.Tensor) -> tuple[torch.Tensor, torch.Tensor, None]:  # type: ignore[override]
+        x, threshold = ctx.saved_tensors
+        bandwidth = ctx.bandwidth
+        x_grad = 0.0 * grad_output  # We don't apply STE to x input
```
chanind, line 53: I think it's fine to just return `None` for the `x_grad` rather than multiplying by 0. I know this is just from the example code

chanind (commented):
Some minor comments but core logic looks great! I'll try training a JumpReLU SAE to make sure this works.

It feels odd that the `l1_coefficient` is used for the JumpReLU sparsity loss coefficient, and that it gets logged as `l1_loss`. Maybe in a follow-up PR we should rename `l1_coefficient` to `sparsity_coefficient` instead? I added a PR https://github.com/jbloomAus/SAELens/pull/357 which lets the forward pass log any losses it wants, so if that gets merged, it would probably make sense to call the loss `sparsity_loss` since there's not technically any L1 loss in jumprelu.

sae_lens/config.py
```diff
@@ -162,6 +163,7 @@ class LanguageModelSAERunnerConfig:
     seed: int = 42
     dtype: str = "float32"  # type: ignore #
     prepend_bos: bool = True
+    threshold: float = 0.001
```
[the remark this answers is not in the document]
anthonyduong9: Oops, thanks for catching that bandwidth wasn't configurable. I've changed the first name's field and added the second field.

sae_lens/training/training_sae.py
anthonyduong9: Oh cool, I've made the change.

anthonyduong9:
> Some minor comments but core logic looks great! I'll try training a JumpReLU SAE to make sure this works.
> 
> It feels odd that the `l1_coefficient` is used for the JumpReLU sparsity loss coefficient, and that it gets logged as `l1_loss`. Maybe in a follow-up PR we should rename `l1_coefficient` to `sparsity_coefficient` instead? I added a PR #357 which lets the forward pass log any losses it wants, so if that gets merged, it would probably make sense to call the loss `sparsity_loss` since there's not technically any L1 loss in jumprelu.

Thanks! Let me know if there's anything else you need from me.

And I'm guessing unless you think we should keep the `l1_coefficient` field for the other architecture(s), and add a `sparsity_coefficient` that's just used for JumpReLU, your suggestion makes sense.

chanind:
Trying a test-run now, but one more thing is that the typing for `architecture` for `LanguageModelSAERunnerConfig` needs to be updated as well - it's currently set to `"standard" | "gated"` so throws a typing error when setting `"jumprelu"`.

chanind:
It seems like the general paradigm of most of the recent SAE architectures is having some sort of sparsity-inducing loss, either l1 in normal SAEs or l0 in jumprelu (or nothing for topk), and then an optional auxiliary loss. I'd be for just calling it `sparsity_coefficient` to handle the architecture differences, but should be discussed in another follow-up issue probably.

chanind:
Screenshot 2024-11-03 at 21 11 48

I tried training for 1B tokens with sparsity coeff of 1e-3 (which I think is reasonable based on the gemma scope paper). Looks like L0 is still coming down by the end of training (Gemma Scope trains on 4B tokens), so I'm assuming this would plateau at a reasonable loss if I kept going.

chanind (approved):
LGTM! Great work with this!

anthonyduong9:
> LGTM! Great work with this!

Thanks!

PR #355, opened 2024-10-30
fix: normalize decoder bias in fold_norm_scaling_factor

### LINKED ISSUE #354: [Bug Report] fold_activation_norm_scaling_factor missing division of decoder bias
If you are submitting a bug report, please fill in the following details and use the tag [bug].

**Describe the bug**
Folding activation scaling factor at the end of SAE training is implemented incorrectly and does not give the same reconstruction ability when loaded after training as it does during training.
[Link to code](https://github.com/jbloomAus/SAELens/blob/main/sae_lens/sae.py#L506-L514). This is caused by the fact that it is not normalizing the decoder bias.

**Fix**
This can be fixed by adding the following line to the folding code:
```
def fold_activation_norm_scaling_factor(
        self, activation_norm_scaling_factor: float
    ):
        self.W_enc.data = self.W_enc.data * activation_norm_scaling_factor
        # previously weren't doing this.
        self.W_dec.data = self.W_dec.data / activation_norm_scaling_factor
        #This is my new line
        self.b_dec.data = self.b_dec.data / activation_norm_scaling_factor
        # once we normalize, we shouldn't need to scale activations.
        self.cfg.normalize_activations = "none"
```

**Additional context**
I have checked the math and this is the right way to do the folding, whether we're subtracting decoder bias from the input or not.  Doing this division post-hoc also fixed my issues with loading trained SAEs.

### Checklist

- [x ] I have checked that there is no similar [issue](https://github.com/jbloomAus/SAELens/issues) in the repo (**required**)

chanind:
# Description

This PR implements the fix specified in #354 and updates tests.

Fixes #354

## Type of change

# Checklist:

### DIFF
--- a/sae_lens/sae.py
+++ b/sae_lens/sae.py
@@ -509,6 +509,7 @@ class SAE(HookedRootModule):
         self.W_enc.data = self.W_enc.data * activation_norm_scaling_factor
         # previously weren't doing this.
         self.W_dec.data = self.W_dec.data / activation_norm_scaling_factor
+        self.b_dec.data = self.b_dec.data / activation_norm_scaling_factor
 
         # once we normalize, we shouldn't need to scale activations.
         self.cfg.normalize_activations = "none"
--- a/tests/unit/training/test_sae_basic.py
+++ b/tests/unit/training/test_sae_basic.py
@@ -107,11 +107,15 @@ def test_sae_fold_w_dec_norm(cfg: LanguageModelSAERunnerConfig):
     torch.testing.assert_close(sae_out_1, sae_out_2)
 
 
+@torch.no_grad()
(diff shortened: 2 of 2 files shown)


curt-tigges:
@chanind Do we know what the scale of the impact of this bug might have been?

chanind:
Seems like it would only be an issue for SAEs trained with `normalize_activations='expected_average_only_in'`, which isn't the default. I'm struggling to get tests to pass with this fix though, the final assertion fails with this fix

chanind:
Looks like the issue was just with the way tests were scaling the output, should be fixed now.

chanind:
Merging as it's currently broken in prod, and seems pretty had to argue with. Thanks @tuomaso!

PR #356, opened 2024-10-31
feat: Add linear probe trainer

tom-pollak:
# Description

Created linear probe trainer which can take a hf dataset (created like in #321) and train a linear probe (provided we have labels for the activations)

By default this can make a pretty wandb run with all your stats, similar to training the SAE.

Screenshot 2024-10-31 at 10 23 39

---

## Type of change

# Checklist:

### DIFF
--- a/sae_lens/analysis/train_probe.py
+++ b/sae_lens/analysis/train_probe.py
@@ -1,0 +1,125 @@
+import pickle
+from datetime import datetime
+from pathlib import Path
+from typing import Any, Dict, Literal, Optional, Tuple
+
+import numpy as np
+import wandb
+import wandb.sklearn
+from datasets import DatasetDict
+from sklearn.linear_model import LogisticRegression
+from sklearn.metrics import classification_report
+from tqdm.auto import tqdm
+
+
+class ProbeTrainer:
+    def __init__(
+        self,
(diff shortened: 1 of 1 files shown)


chanind:
I would personally view this as out-of-scope of this library, since there's nothing SAE-specific about training a linear probe. Curious to hear others' thoughts though!

hijohnnylin:
> I would personally view this as out-of-scope of this library, since there's nothing SAE-specific about training a linear probe. Curious to hear others' thoughts though!

FWIW, I'm a +1 to including this in SAELens, and/or some repo similarly easily accessible. possibly the scope of SAELens should be widened. I would prefer not to have a separate library for TranscoderLens, CrosscoderLens, ProbeLens etc. Though I defer to you, @curt-tigges, etc.

curt-tigges:
We've had some discussion internally about creating and maintaining a probe-specific library, which I think would be a better fit than SAELens for this. I think one of the main benefits of having a library is having common code etc. but probing is rather independent and there isn't really anything shared with the rest of the code. There's not really a specific reason to have it in SAELens.

Proposal: I'm inclined to just go ahead and create a sketch for an initial probing library and ask @tom-pollak to resubmit this code to that. @jbloomAus any thoughts?

PR #357, opened 2024-11-03
chore: more flexible training losses

chanind:
# Description

This PR refactors the `TrainStepOutput` in `TrainingSAE.training_forward_pass()` to allow passing `losses` as a dict rather than a hardcoded set of specific losses. The current implementation requires always providing values for ghost grad loss and auxiliary loss, even when it doesn't make sense for the training method. As SAELens supports more architectures, there will likely be more losses like this that only make sense for certain architectures, so it makes sense to let the train forward step define the losses that make sense for the circumstances.

This should also make it easier for researchers to hack on the SAELens SAEs since it means they can define any losses they want to track.

This PR also allows the losses to remain as tensors rather than being floats. The current implementation requires calling `.item()` on each loss, which forces a synchronization with the GPU and likely slows down training slightly. We only ever use the losses for logging, which happens by default once every 100 steps, so this synchronization is unnecessary.

## Type of change

- [x] Refactor

# Checklist:

### DIFF
--- a/sae_lens/training/sae_trainer.py
+++ b/sae_lens/training/sae_trainer.py
@@ -402,9 +402,11 @@ class SAETrainer:
     def _update_pbar(self, step_output: TrainStepOutput, pbar: tqdm, update_interval: int = 100):  # type: ignore
 
         if self.n_training_steps % update_interval == 0:
-            pbar.set_description(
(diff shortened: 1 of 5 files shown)


chanind:
discussed / approved offline

PR #361, opened 2024-11-05
fix: load the same config from_pretrained and get_sae_config

### LINKED ISSUE #351: [Bug Report] get_sae_config doesn't return the same config as SAE.from_pretrained due to defaulting
**Describe the bug**
1) get `cfg_dict` using `SAE.from_pretrained`
2) get the config for the same SAE using `get_sae_config`
3) outputs are different for 1 and 2
See sample code.

Reason seems to be that there are defaults that are added through `from_pretrained` that aren't added for `get_sae_config`: 
https://github.com/jbloomAus/SAELens/blob/6d6eaef343fd72add6e26d4c13307643a62c41bf/sae_lens/toolkit/pretrained_sae_loaders.py#L122-L131

IMO the two outputs should be the same (either without the defaults or with the defaults). I'm a little iffy about the `handle_config_defaulting` in general, because it seems to assume things that may not be true. For example, it assumes that any config that doesn't have `prepend_bos` means that prepend_bos was True. That seems like a pretty big assumption?

**Code example**
```
from pprint import pprint
from sae_lens import SAE
from sae_lens.toolkit.pretrained_sae_loaders import (
    SAEConfigLoadOptions,
    get_sae_config,
)

sae, cfg_dict, sparsity = SAE.from_pretrained(release="gpt2-small-res-jb", sae_id="blocks.0.hook_resid_pre")
pprint(cfg_dict)

get_sae_cfg_output = get_sae_config(
    "gpt2-small-res-jb",
    "blocks.0.hook_resid_pre",
    SAEConfigLoadOptions()
)

pprint(get_sae_cfg_output)

# Compare dictionaries by finding keys in cfg_dict that aren't in cfg and vice versa
missing_in_cfg = set(cfg_dict.keys()) - set(get_sae_cfg_output.keys())
missing_in_cfg_dict = set(get_sae_cfg_output.keys()) - set(cfg_dict.keys())

print("\nKeys in cfg_dict but not in get_sae_cfg_output:")
pprint(missing_in_cfg)

print("\nKeys in cfg but not in cfg_dict:")
pprint(missing_in_cfg_dict)
```

### Checklist

chanind:
# Description

This PR unifies the code path for `from_pretrained` and `get_sae_config` so the config dict should match, and adds a test asserting this for a jbloom/gpt2 SAE.

Fixes #351

## Type of change

# Checklist:

### DIFF
--- a/sae_lens/sae.py
+++ b/sae_lens/sae.py
@@ -620,9 +620,6 @@ class SAE(HookedRootModule):
             )
         sae_info = sae_directory.get(release, None)
         config_overrides = sae_info.config_overrides if sae_info is not None else None
-        neuronpedia_id = (
-            sae_info.neuronpedia_id[sae_id] if sae_info is not None else None
-        )
 
         conversion_loader_name = get_conversion_loader_name(sae_info)
         conversion_loader = NAMED_PRETRAINED_SAE_LOADERS[conversion_loader_name]
@@ -637,7 +634,6 @@ class SAE(HookedRootModule):
 
         sae = cls(SAEConfig.from_dict(cfg_dict))
         sae.load_state_dict(state_dict)
-        sae.cfg.neuronpedia_id = neuronpedia_id
 
         # Check if normalization is 'expected_average_only_in'
         if cfg_dict.get("normalize_activations") == "expected_average_only_in":
--- a/sae_lens/toolkit/pretrained_sae_loaders.py
+++ b/sae_lens/toolkit/pretrained_sae_loaders.py
@@ -474,7 +474,8 @@ def get_sae_config(
     repo_id, folder_name = get_repo_id_and_folder_name(release, sae_id=sae_id)
     cfg_overrides = options.cfg_overrides or {}
     if sae_info is not None:
-        sae_info_overrides: dict[str, Any] = sae_info.config_overrides or {}
+        # avoid modifying the original dict
+        sae_info_overrides: dict[str, Any] = {**(sae_info.config_overrides or {})}
         if sae_info.neuronpedia_id is not None:
             sae_info_overrides["neuronpedia_id"] = sae_info.neuronpedia_id.get(sae_id)
         cfg_overrides = {**sae_info_overrides, **cfg_overrides}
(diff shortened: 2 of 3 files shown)


hijohnnylin:
thanks!

i tested using my repro code from the issue and it appears to be fixed. you are good to merge.

only thing i noticed during review was that there's a `neuronpedia` property and also a `neuronpedia_id` property, both which seem to always return None even if the neuronpedia value is set in the yaml file: https://github.com/jbloomAus/SAELens/blob/84203cde704f56088cb4437233d02cba5e5f0dd9/sae_lens/toolkit/pretrained_sae_loaders.py#L129 and https://github.com/jbloomAus/SAELens/blob/84203cde704f56088cb4437233d02cba5e5f0dd9/sae_lens/toolkit/pretrained_sae_loaders.py#L453

let me know if i should open a separate issue for that, happy to do it.

hijohnnylin (approved):
thanks!

i tested using my repro code from the issue and it appears to be fixed. you are good to merge.

only thing i noticed during review was that there's a `neuronpedia` property and also a `neuronpedia_id` property, both which seem to always return None even if the neuronpedia value is set in the yaml file: https://github.com/jbloomAus/SAELens/blob/84203cde704f56088cb4437233d02cba5e5f0dd9/sae_lens/toolkit/pretrained_sae_loaders.py#L129 and https://github.com/jbloomAus/SAELens/blob/84203cde704f56088cb4437233d02cba5e5f0dd9/sae_lens/toolkit/pretrained_sae_loaders.py#L453

let me know if i should open a separate issue for that, happy to do it.

chanind:
I noticed that as well - it looks like the neuronpedia_id property is only set on the config in the SAE, but not in the config_dict that gets returned from `SAE.from_pretrained()`. I'm not sure what the reason for that is, or if it's simply an oversight. I can add it to the the config_dict so it's set every if that makes more sense. I'm also not sure what the idea with the `neuronpedia` property is

chanind:
I updated the PR to include `neuronpedia_id` in the config in both places. I think the `neuronpedia` properly was likely just a typo, since there's no equivalent property on the SAEConfig class.

PR #367, opened 2024-11-10
Faster cleanup of sharded datasets

tom-pollak:
# Description

Builds upon #321. Previously I used `dataset.save_to_disk` to write the final dataset, but this will rewrite the entire dataset to disk, which is very slow. Instead I manually move the shards to the standard hf format which allows us not to resave the entire dataset.

## Type of change

- [X] Performance improvement

This should be externally identical, but just saving the dataset faster. My previous cached activation runner tests validate the change well.

Benchmarking cached activation benchmark on my mac:

```
poetry run py.test tests/benchmark/test_cache_activations_runner.py::test_cache_activations_runner -s

Old: Caching activations took: 81.5707

New: Caching activations took: 68.0761
```

I expect this to be even more pronounced for larger runs, where computing activations scales faster than disk speed.

# Checklist:

### DIFF
--- a/sae_lens/cache_activations_runner.py
+++ b/sae_lens/cache_activations_runner.py
@@ -1,13 +1,14 @@
 import io
 import json
 import math
-import os
 import shutil
 from dataclasses import asdict
+from pathlib import Path
 
 import einops
 import torch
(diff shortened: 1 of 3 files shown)


chanind:
Nice, so this uses less disk space as well? Looks good!

PR #370, opened 2024-11-11
feat: Topk SAE training

### LINKED ISSUE #202: [Proposal] Implement OpenAI's TopK Auxilliary for preventing dead latents.
### Proposal 

Support the auxiliary loss function for avoiding dead neurons in top-k architectured SAEs described [here](https://cdn.openai.com/papers/sparse-autoencoders.pdf).

### Motivation

We support top-K SAEs as of #201 but haven't implemented the auxiliary loss. 

### Pitch

Implement the loss. Hopefully it's not too messy. We are probably due for some refactoring anyway but I expect this shouldn't be required to get this working. 

> We find two important ingredients for preventing dead latents: we initialize the encoder to the transpose of the
decoder, and we use an auxiliary loss that models reconstruction error using the top-kaux dead latents
(see Section A.2 for more details). Using these techniques, even in our largest (16 million latent)
autoencoder only 7% of latents are dead.

Screenshot 2024-07-01 at 4 28 39 PM

### Alternatives

NA

### Additional context

- I'd expect some benchmark runs that show the training works. Can compare to stats OAI gets on GPT2 small to show it's working.
- This loss will cost us some training performance but there's no alternative. 
- Config wise I think it might make sense to move from `use_ghost_grads` being a bool to it being a Literal and having a value for this procedure (which is only valid if it's a top-K SAE). We can handle loading old configs in the pretrained SAE loading code to maintain backward compatibility with saved configs. 

### Checklist

chanind:
# Description

This PR implements topk SAE training by adding the topk auxiliary loss. This PR makes several design choices:

- `topk` is specified as an `architecture` for training, so it's similar to `gated` and `jumprelu` SAEs. This seems fit the idea of an SAE architecture since it has its own custom training routine and losses, and it seems strange to call "jumprelu" and "gated" architectures, but not topk.
- This PR is based heavily on https://github.com/EleutherAI/sae/blob/main/sae/sae.py#L191 rather than the original topk SAEs paper. The Eleuther version requires no additional config options for the auxiliary loss, which means less changes to our existing config, while allowing us to add those config options later if we want. We may also want to refactor the way our config / SAE classes work now that we support so many different architectures anyway.
- This PR makes the following changes to Eleuther's topk aux loss:
    - Removes the variance normalization. Eleuther divides all their losses by variance, but we don't do that anywhere, so I removed this normalization to be consistent
    - Take the mean loss over batches. Eleuther takes the sum, but in our other losses we take a mean over batches rather than sum over batches, so I changed this to be consistent with that.

Our implementation of topk training is likely less efficient than Eleuther's as they use a custom sparse kernel for the SAE decoder (see https://github.com/EleutherAI/sae/blob/main/sae/kernels.py). We can try to support something like this in the future, but it will likely require a bit of refactoring before we can support a special decoder kernel just for topk.

I'm currently running some test training runs to make sure things look decent, and will upgrade this PR from draft when those are complete.

Fixes #202

training test run dashboard: https://api.wandb.ai/links/chanind/zju8dl70

## Type of change

# Checklist:


sae_lens/training/sae_trainer.py
```diff
@@ -114,13 +114,13 @@ def __init__(
...
         self.l1_scheduler = L1Scheduler(
-            l1_warm_up_steps=cfg.l1_warm_up_steps,  # type: ignore
+            l1_warm_up_steps=cfg.l1_warm_up_steps,
             total_steps=cfg.total_training_steps,
             final_l1_coefficient=cfg.l1_coefficient,
         )
 
         # Setup autocast if using
-        self.scaler = torch.cuda.amp.GradScaler(enabled=self.cfg.autocast)
+        self.scaler = torch.amp.GradScaler(device="cuda", enabled=self.cfg.autocast)
```
curt-tigges, line 124: This should probably use whatever device is set through the config, rather than being hardcoded to CUDA.

tests/unit/training/test_training_sae.py
```diff
@@ -4,7 +4,11 @@
 import torch
 
 from sae_lens.sae import SAE
-from sae_lens.training.training_sae import TrainingSAE, TrainingSAEConfig
+from sae_lens.training.training_sae import (
+    TrainingSAE,
+    TrainingSAEConfig,
+    _calculate_topk_aux_acts,
```
curt-tigges, line 10: If this function is intended to be imported by other files/classes, probably should not be an underscore/internal function.

curt-tigges (commented):
Looks good other than the two comments I made.

tests/unit/training/test_training_sae.py
chanind: It's only imported by its own unit test, it should not be imported externally which is why it has an underscore

PR #372, opened 2024-11-11
chore: add codecov.yaml and exclude legacy files

chanind:
# Description

We have a number of legacy files which we may want to delete, but at least we shouldn't be including them in code coverage stats as it makes the project look like it's not properly tested. This PR adds a codecov.yaml where we can further configure codecov, and excludes the following files:

  - sae_lens/analysis/feature_statistics.py
  - sae_lens/training/tsea.py
  - sae_lens/toolkit/pretrained_saes.py

These files can probably be deleted in the future, they're not documented or referenced anywhere except some legacy jupyter notebooks.

## Type of change

- [x] Chore


chanind:
merging as this is just affects codecov stats, which should be pretty non-controversial

PR #373, opened 2024-11-11
fix: fixing jumprelu encode and save/load

chanind:
# Description

This PR fixes some minor issues related to jumprelu SAEs:
- fixes a bug where we were accidentally creating a `threshold` param and a `log_threshold` param on `TrainingSAE`, and calling `encode()` incorrectly used the `threshold` while `encode_with_hidden_pre` correctly used the `log_threshold`
- enables loading saved jumprelu SAEs for further training
- fixes a bug where jumprelu bandwidth and init threshold weren't being saved along with the rest of the config

## Type of change

# Checklist:

### DIFF
--- a/sae_lens/training/training_sae.py
+++ b/sae_lens/training/training_sae.py
@@ -266,29 +266,9 @@ class TrainingSAE(SAE):
     def initialize_weights_jumprelu(self):
         # same as the superclass, except we use a log_threshold parameter instead of threshold
         self.log_threshold = nn.Parameter(
-            torch.ones(self.cfg.d_sae, dtype=self.dtype, device=self.device)
+            torch.empty(self.cfg.d_sae, dtype=self.dtype, device=self.device)
         )
-        self.b_enc = nn.Parameter(
-            torch.zeros(self.cfg.d_sae, dtype=self.dtype, device=self.device)
-        )
-
-        self.W_dec = nn.Parameter(
-            torch.nn.init.kaiming_uniform_(
-                torch.empty(
-                    self.cfg.d_sae, self.cfg.d_in, dtype=self.dtype, device=self.device
-                )
-            )
-        )
-        self.W_enc = nn.Parameter(
-            torch.nn.init.kaiming_uniform_(
-                torch.empty(
-                    self.cfg.d_in, self.cfg.d_sae, dtype=self.dtype, device=self.device
-                )
-            )
-        )
-        self.b_dec = nn.Parameter(
-            torch.zeros(self.cfg.d_in, dtype=self.dtype, device=self.device)
-        )
+        self.initialize_weights_basic()
 
     @property
     def threshold(self) -> torch.Tensor:
--- a/tests/unit/training/test_training_sae.py
+++ b/tests/unit/training/test_training_sae.py
@@ -4,7 +4,7 @@ import pytest
 import torch
 
 from sae_lens.sae import SAE
-from sae_lens.training.training_sae import JumpReLU, TrainingSAE, TrainingSAEConfig
+from sae_lens.training.training_sae import TrainingSAE, TrainingSAEConfig
 from tests.unit.helpers import build_sae_cfg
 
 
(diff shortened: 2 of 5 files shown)


chanind:
Informally tagging @anthonyduong9 as a reviewer

sae_lens/training/training_sae.py
```diff
@@ -263,6 +263,39 @@ def __init__(self, cfg: TrainingSAEConfig, use_error_term: bool = False):
...
+        self.W_enc = nn.Parameter(
+            torch.nn.init.kaiming_uniform_(
+                torch.empty(
+                    self.cfg.d_in, self.cfg.d_sae, dtype=self.dtype, device=self.device
+                )
+            )
+        )
+        self.b_dec = nn.Parameter(
+            torch.zeros(self.cfg.d_in, dtype=self.dtype, device=self.device)
+        )
```
anthonyduong9, line 291: This block is also in `SAE.initialize_weights_basic()` and `SAE.initialize_weights_jumprelu()` (where we have a comment about the parameters being the same as in the former). Extracting it to a function (e.g. a method on `SAE`) or potentially refactoring so we can just call `SAE.initialize_weights_basic()` here and in `SAE.initialize_weights_jumprelu()` would reduce a lot of code and make it self-documenting that the three `initalize_` methods are almost the same. I'm not sure if we should do it in this PR or a separate one though.

tests/unit/training/test_training_sae.py
```diff
@@ -37,3 +40,110 @@ def test_TrainingSAE_training_forward_pass_can_scale_sparsity_penalty_by_decoder
...
+        assert torch.allclose(param, loaded_training_sae.state_dict()[param_name])
+
+    test_input = torch.randn(32, cfg.d_in)
+    training_sae_out = training_sae.encode_with_hidden_pre_fn(test_input)[0]
+    loaded_training_sae_out = loaded_training_sae.encode_with_hidden_pre_fn(test_input)[
+        0
+    ]
+    loaded_sae_out = loaded_sae.encode(test_input)
+    assert torch.allclose(training_sae_out, loaded_training_sae_out)
+    assert torch.allclose(training_sae_out, loaded_sae_out)
```
anthonyduong9, line 149: This test looks like a test that I wrote, except I just realized I accidentally passed `"gated"` instead of `"jumprelu"`.

https://github.com/jbloomAus/SAELens/blob/156ddc9fbdc92217147257886f38a2d719e45bff/tests/unit/training/test_sae_basic.py#L206-L229

So we should delete the test that I wrote too.

tests/unit/training/test_training_sae.py
```diff
@@ -37,3 +40,110 @@ def test_TrainingSAE_training_forward_pass_can_scale_sparsity_penalty_by_decoder
...
+    assert "threshold" not in param_names
+    assert torch.allclose(
+        sae.threshold,
+        torch.ones_like(sae.log_threshold.data) * cfg.jumprelu_init_threshold,
+    )
+
+
+def test_TrainingSAE_jumprelu_sae_encoding():
+    cfg = build_sae_cfg(architecture="jumprelu")
+    sae = TrainingSAE.from_dict(cfg.get_training_sae_cfg_dict())
```
anthonyduong9, line 71: We have this block in three of the tests, so it might help to extract it to a fixture.

tests/unit/training/test_training_sae.py
```diff
@@ -37,3 +40,110 @@ def test_TrainingSAE_training_forward_pass_can_scale_sparsity_penalty_by_decoder
...
+        (torch.pow((train_step_output.sae_out - x.float()), 2))
+        .sum(dim=-1)
+        .mean()
+        .detach()
+        .float()
+    )
+
+    assert (
+        pytest.approx(train_step_output.losses["mse_loss"].item()) == expected_mse_loss  # type: ignore
+    )
```
anthonyduong9, line 126: What's the reasoning for moving this out of `test_jumprelu_sae.py`?

I created that file as I was pattern matching with `test_gated_sae.py` and probably other files but am wondering if we should move tests from `test_gated_sae.py` into here as well (though not in this PR), and generally, how we want to organize tests.

sae_lens/training/training_sae.py
chanind: aah good call - I'll make that change

tests/unit/training/test_training_sae.py
chanind: I'm not a huge fan of pulling out stuff into fixtures unless it saves a lot of lines of code. Making this into a fixture means you can't just look at the test and understand what it's doing, you also need to scroll to look at the fixture. Since this is just 2 lines of code, I'm rather leave it as is IMO

tests/unit/training/test_training_sae.py
chanind: oh my bad I didn't see `test_gated_sae.py` - I would argue that should be in here too. I usually aim to have a 1-to-1 mapping of test file to library file, so if there's `sae_lens/training/training_sae.py` then I would expect a corresponding `tests/training/test_training_sae.py`. I'm not a fan of having tests not map onto source files since it becomes hard to know where tests should go, and where to look to see where something is tested. We should probably write up something about how we expect code / tests to be structured since I don't think this was discussed officially. Regardless, I think we should have some deterministic way to structure tests so it's not arbitrary where things are tested.

chanind: It looks like there's a lot of tests files related to the SAE class that don't follow this paradigm, so I reverted the changes to `test_jumprelu_sae.py` pending a larger discussion

PR #376, opened 2024-11-14
Fix parameter naming: Separate l0_lambda from l1_coefficient

### LINKED ISSUE #360: [Proposal] Rename `l1_coefficient` to `sparsity_coefficient`
### Proposal 

Now that we support JumpReLU training, the `l1_coefficient` is confusing since jumprelu use a l0 loss, not l1, for training. We should rename this parameter to `sparsity_coefficient` since it is a coefficient used to generally promote sparsity. We should also rename `l1_warmup_steps` to `sparsity_warmup_steps`.

### Motivation

It is confusing to see `l1_coefficient` used for JumpReLU training which doesn't use L1 loss.

### Alternatives

Alternatively, we could add a separate `l0_coefficient` / `l0_warmup_steps` which is only used for jumprelu training and error if `l1_coefficient` is provided. This would also potentially allow training a jumprelu with both L0 and L1 loss if desired.

### Checklist

muyo8692:
Resolves #360

## Description
Currently, when using the JumpReLU architecture which involves l0 regularization, the code uses `l1_coefficient` parameter name for the l0 loss calculation. This naming is confusing and makes the configuration less intuitive. This PR introduces a dedicated `l0_lambda` parameter to clearly separate these two regularization terms.

## Changes
- Added new parameter `l0_lambda` in config for l0 loss calculation (default: 0.0)
- Added validation to enforce `l0_lambda` specification when `architecture="jumprelu"` is selected
- Maintained existing behavior: using JumpReLU without specifying `l0_lambda` will raise an error to prevent silent failures

## Testing
- Verified numerical equivalence: models trained with the original code and modified code produce identical metrics
- Added error handling test for missing `l0_lambda` with JumpReLU architecture

## Notes
This change improves code clarity while ensuring backward compatibility is maintained through proper error handling rather than silent parameter reuse.

Screenshot 2024-11-14 at 10 41 13


chanind:
I'm a bit worried about adding lots of architecture-specific parameters to the config as this seems like it will get out of hand pretty quickly. I agree that using `l1_coefficient` for JumpReLU doesn't feel great, but I worry about the confusion burden of having lots of different parameters that apply only to 1 architecture in the config. Longer-term, we should probably move to a config structure with something like a `architecture_params` dictionary in the config where each architecture can have separate settings or something.

If we do decide to add this parameter, I'm not a fan of the name `l0_lambda` since we have `l1_coefficient`, and these feel like they should be consistent.

The benefit of an approach like in #360 where we just have a `sparsity_coefficient` is we don't need to have multiple params and multiple schedulers even though only 1 is actually relevant.

What do other thing? cc @curt-tigges @hijohnnylin @anthonyduong9

.vscode/settings.json
```diff
@@ -6,7 +6,7 @@
   "python.testing.pytestEnabled": true,
 
   "[python]": {
-    "editor.defaultFormatter": "ms-python.black-formatter",
+    "editor.defaultFormatter": "charliermarsh.ruff",
```
chanind, line 9: This looks like a mistake - we don't use ruff currently

sae_lens/training/optim.py
```diff
@@ -157,3 +157,61 @@ def load_state_dict(self, state_dict: dict[str, Any]):
         """Loads all state apart from attached SAE."""
         for k in state_dict:
             setattr(self, k, state_dict[k])
+
+
+class L0Scheduler:
```
chanind, line 162: This feels very heavy-weight to add another class entirely that's basically identical to `L1Scheduler`. We should just make the `L1Scheduler` into `CoefficientScheduler` or something so it can be used in multiple places

muyo8692:
> I'm a bit worried about adding lots of architecture-specific parameters to the config as this seems like it will get out of hand pretty quickly. I agree that using `l1_coefficient` for JumpReLU doesn't feel great, but I worry about the confusion burden of having lots of different parameters that apply only to 1 architecture in the config. Longer-term, we should probably move to a config structure with something like a `architecture_params` dictionary in the config where each architecture can have separate settings or something.
> 
> If we do decide to add this parameter, I'm not a fan of the name `l0_lambda` since we have `l1_coefficient`, and these feel like they should be consistent.
> 
> The benefit of an approach like in #360 where we just have a `sparsity_coefficient` is we don't need to have multiple params and multiple schedulers even though only 1 is actually relevant.
> 

thanks for the comments & totally agree with add `sparsity_coefficient` instead of individual configs.
I should have time to work on it this weekend

muyo8692:
@chanind I've made the adjustment, it should work fine now

chanind:
Thank you for doing this! There's a few more complications though:
- @curt-tigges might be refactoring the configs / SAE classes soon to have separate architecture-specific config sections, which would make this unnecessary. @curt-tigges can talk to the timeline for that
- We should make the old options still work for backwards compatibility with a deprecation message instructing the user to use `sparsity_coefficient` if their config has `l1_coefficient` still. Otherwise this would be a breaking change.
- Docs/notebooks also need to be updated to talk about `sparsity_coefficient` instead of `l1_coefficient`. This is probably just a find/replace.

I'm not sure if it's worth it though if Curt is going to be refactoring this soon anyway. Up to you if you think it's worth doing!

muyo8692:
Good to hear there will be a refactoring of the configs, in this case I think I'll just wait for the new version! closing the PR

PR #388, opened 2024-11-25
Replace print with logging

### LINKED ISSUE #43: Use logging instead of print in codebase
Currently, logging info is output using `print()`, but this makes it hard to control the verbosity level of the output, and cannot distinguish between debugging / info statements and warnings / errors. For instance, in [sae_training/config.py](https://github.com/jbloomAus/mats_sae_training/blob/main/sae_training/config.py#L117-L155), creating a config object prints a lot of statements, but these are a mix of debugging / info / warnings, and would be better suited to using the Python [logging module](https://docs.python.org/3/library/logging.html).

anthonyduong9:
# Description

Replaces `print()` with logging methods called on a logger, where we don't display console output for a command line script, and excluding notebooks and the `benchmark/tests` directory. This allows us to track the severity of events and makes it possible to configure the logging verbosity. Currently, we're just using `print()`.

Fixes #43

## Type of change

# Checklist:

### Performance Check.

- [ ] L0
- [ ] CE Loss
- [ ] MSE Loss

### DIFF
--- a/sae_lens/__init__.py
+++ b/sae_lens/__init__.py
@@ -1,5 +1,8 @@
 __version__ = "4.4.5"
 
+import logging
+
+logger = logging.getLogger(__name__)
 
 from .analysis.hooked_sae_transformer import HookedSAETransformer
 from .cache_activations_runner import CacheActivationsRunner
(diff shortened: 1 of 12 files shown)


chanind:
LGTM! Just merged main and resolved conflicts, will merge once CI passes

PR #389, opened 2024-11-25
Cleaned up CacheActionsRunnerConfig

tom-pollak:
# Description

Before `CacheActivationConfig` had a inconsistent config file for some interopability with `LanguageModelSAERunnerConfig`. It was kind of unclear which parameters were necessary vs redundant, and just was fairly unclear.

Simplified to the required arguments:

- `hf_dataset_path`: Tokenized or untokenized dataset
- `total_training_tokens`
- `model_name`
- `model_batch_size`
- `hook_name`
- `final_hook_layer`
- `d_in`

I think this scheme captures everything you need when attempting to cache activations and makes it a lot easier to reason about.

Optional:

```
activation_save_path # defaults to "activations/{dataset}/{model}/{hook_name}
shuffle=True
prepend_bos=True
streaming=True
seqpos_slice
buffer_size_gb=2 # Size of each buffer. Affects memory usage and saving freq
device="cuda" or "cpu"
dtype="float32"
autocast_lm=False
compile_llm=True
hf_repo_id # Push to hf
model_kwargs # `run_with_cache`
model_from_pretrained_kwargs
```

Fixes # (issue)

## Type of change

- [X] Breaking change (fix or feature that would cause existing functionality to not work as expected)

# Checklist:

### DIFF
--- a/sae_lens/config.py
+++ b/sae_lens/config.py
@@ -2,7 +2,6 @@ import json
 import math
 import os
 from dataclasses import dataclass, field
-from pathlib import Path
 from typing import Any, Literal, Optional, cast
 
 import torch
@@ -489,16 +488,16 @@ class CacheActivationsRunnerConfig:
     Configuration for creating and caching activations of an LLM.
 
     Args:
-        hf_dataset_path (str): The path to the Hugging Face dataset. This may be tokenized or not.
+        dataset_path (str): The path to the Hugging Face dataset. This may be tokenized or not.
         model_name (str): The name of the model to use.
         model_batch_size (int): How many prompts are in the batch of the language model when generating activations.
         hook_name (str): The name of the hook to use.
-        final_hook_layer (int): The layer of the final hook. Currently only support a single hook, so this should be the same as hook_name.
+        hook_layer (int): The layer of the final hook. Currently only support a single hook, so this should be the same as hook_name.
         d_in (int): Dimension of the model.
         total_training_tokens (int): Total number of tokens to process.
         context_size (int): Context size to process. Can be left as -1 if the dataset is tokenized.
         model_class_name (str): The name of the class of the model to use. This should be either `HookedTransformer` or `HookedMamba`.
-        activation_save_path (str, optional): The path to save the activations.
+        new_cached_activations_path (str, optional): The path to save the activations.
         shuffle (bool): Whether to shuffle the dataset.
         seed (int): The seed to use for shuffling.
         dtype (str): Datatype of activations to be stored.
@@ -519,22 +518,18 @@ class CacheActivationsRunnerConfig:
         dataset_trust_remote_code (bool): Whether to trust remote code when loading datasets from Huggingface.
     """
 
-    hf_dataset_path: str
-
+    dataset_path: str
     model_name: str
     model_batch_size: int
-
     hook_name: str
-    final_hook_layer: int  # Layer of final hook. Same layer as hook_name
+    hook_layer: int
     d_in: int
-
-    total_training_tokens: int
+    training_tokens: int
 
     context_size: int = -1  # Required if dataset is not tokenized
     model_class_name: str = "HookedTransformer"
     # defaults to "activations/{dataset}/{model}/{hook_name}
-    activation_save_path: str | None = None  # type: ignore
-
+    new_cached_activations_path: str | None = None
     shuffle: bool = True
     seed: int = 42
     dtype: str = "float32"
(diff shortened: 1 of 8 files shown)


sae_lens/config.py
```diff
@@ -478,76 +486,136 @@ def from_json(cls, path: str) -> "LanguageModelSAERunnerConfig":
...
+
+    # Activation Store
+    prepend_bos: bool = True
+    seqpos_slice: tuple[int | None, ...] = (None,)
+    streaming: bool = True
+    autocast_lm: bool = False
+    dataset_trust_remote_code: bool | None = None
+
+    # set in __post_init__
+    tokens_in_buffer: int = -1
```
chanind, line 564: nit: this would be better as `@property` functions rather than direct config params

sae_lens/config.py
```diff
@@ -478,76 +486,136 @@ def from_json(cls, path: str) -> "LanguageModelSAERunnerConfig":
...
-    hook_head_index: Optional[int] = None
-    dataset_path: str = ""
-    dataset_trust_remote_code: bool | None = None
-    streaming: bool = True
-    is_dataset_tokenized: bool = True
-    context_size: int = 128
-    new_cached_activations_path: Optional[str] = (
-        None  # Defaults to "activations/{dataset}/{model}/{full_hook_name}_{hook_head_index}"
-    )
+    hf_dataset_path: str
```
chanind, line 522: this looks like a breaking change. Do we need to change this name? This would require a major version bump and would be annoying for anyone currently using the library as they will need to change their configs

sae_lens/config.py
```diff
@@ -478,76 +486,136 @@ def from_json(cls, path: str) -> "LanguageModelSAERunnerConfig":
...
-    hf_is_private_repo: bool = False
+    model_name: str
+    model_batch_size: int
 
-    # dont' specify this since you don't want to load from disk with the cache runner.
-    cached_activations_path: Optional[str] = None
-    # SAE Parameters
-    d_in: int = 512
+    hook_name: str
+    final_hook_layer: int  # Layer of final hook. Same layer as hook_name
```
chanind, line 528: This also looks like a break change. Why is this renamed?

sae_lens/training/activations_store.py
```diff
@@ -49,13 +49,77 @@ class ActivationsStore:
     _storage_buffer: torch.Tensor | None = None
     device: torch.device
 
+    @classmethod
+    def _from_save_activations(
+        cls,
+        model: HookedRootModule,
+        cfg: CacheActivationsRunnerConfig,
+    ) -> "ActivationsStore":
```
chanind, line 57: Why is this different than `from_cache_activations` below? These both take a cache activations runner config and return a store

sae_lens/training/activations_store.py
```diff
@@ -563,15 +627,17 @@ def get_buffer(
                 total_size, context_size, num_layers, d_in, raise_on_epoch_end
             )
 
-        refill_iterator = range(0, batch_size * n_batches_in_buffer, batch_size)
+        refill_iterator = range(0, total_size, batch_size)
```
chanind, line 630: 👍

sae_lens/config.py
```diff
@@ -478,76 +486,136 @@ def from_json(cls, path: str) -> "LanguageModelSAERunnerConfig":
...
+    d_in: int
 
-    # Activation Store Parameters
-    n_batches_in_buffer: int = 20
-    training_tokens: int = 2_000_000
-    store_batch_size_prompts: int = 32
-    train_batch_size_tokens: int = 4096
-    normalize_activations: str = "none"  # should always be none for activation caching
-    seqpos_slice: tuple[int | None, ...] = (None,)
+    total_training_tokens: int
```
chanind, line 531: This also looks like a breaking change

chanind (commented):
It looks like this has some nice refactoring, but also a bunch of breaking changes. Do these params need to be renamed? It's nice if the params match what's in the other runner config as much as possible

tom-pollak:
sorry about that! renamed accordingly. what do you think about renaming `store_batch_size_prompts` to `model_batch_size`? I could keep `store_batch_size_prompts` as a deprecated param

Summary of changes to config:
```
- buffer_size_gb replaces n_batches_in_buffer

Removed:
- hook_head_index
- cached_activations_path
- is_dataset_tokenized 
- train_batch_size_tokens
- normalize_activations
- act_store_device
```

Technically shouldn't it still be breaking since we removed some parameters? or I could add a `**kwargs` to catch these

chanind:
I'd rather not use `**kwargs` since that would mean silently ignoring options. I'll just bump the major version, but it looks like for most users it should require fewer changes this time.

PR #391, opened 2024-11-25
Move from manual sharding to HF dataset builder.

tom-pollak:
# Description

Depends on #389.

Inspired by:
https://opensourcemechanistic.slack.com/archives/C07EHMK3XC7/p1732413633220709

Instead of manually writing the single arrow shards, we can create a
dataset builder that can do this more efficiently. This speeds up saving
quite a lot, old method spent a some time calculating the fingerprint of
the shard, which was unecessary and would require a hack to get around.

> Along with this change, I also switched to a 1D activation scheme.

- Previously the dataset was stored as a `(seq_len d_in)` array.
- Now stored as a flat `d_in`

Primary reason for this change is shuffling activations. I found that by
using activations sequence, the activations are not properly shuffled.
This is a problem with `ActivationCache` too but there's not a great
solution for it there.

You can observe this in the loss of the SAE by using small buffer sizes
with either using cache or `ActivationStore`.

Fixes # (issue)

## Type of change

- [X] Breaking change `(d_in)` activations vs `(seq_len, d_in)`

# Checklist:


tom-pollak:
Old
```
No saving took: 28.3446
HF Dataset took: 31.2326 ****                                                                                                                                                                                                                                                                                                                                                                                                                             
HF Dataset size: 636.83 MB
Safetensors took: 28.4925                                                                                                                                                                                                                                                                                                                                                                                                                             
Safetensors size: 635.50 MB
```

New
```
No saving took: 28.3992                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                         
HF Dataset took: 28.5294 ****                                                                                                                                                                                                                                                                                                                                                                                                                              
HF Dataset size: 635.56 MB
Safetensors took: 28.6687                                                                                                                                                                                                                                                                                                                                                                                                                             
Safetensors size: 635.50 MB
```

Not a huge amount of difference in benchmarks, slightly faster.

PR #395, opened 2024-12-02
Save estimated norm scaling factor during checkpointing

### LINKED ISSUE #381: [Bug Report] fold_activation_norm_scaling_factor should be performed before saving all checkpoints, not only the final checkpoint
If using `normalize_activations` = `expected_average_only_in`, we scale activations by a scaling factor being training the SAE. Without calling `fold_activation_norm_scaling_factor()`, the SAE will not work properly with unscaled activations. We do not save the scaling factor along with the SAE, so there's no way to use an SAE that's been saved without the activation scaling also being saved or folded in.

SAELens trainer calls  `fold_activation_norm_scaling_factor()` before saving the final checkpoint, but this is not called before intermediate checkpoints. This means the intermediate checkpoints are not usable SAEs. 

We should either call `fold_activation_norm_scaling_factor()` before saving a checkpoint, or include the norm scaling factor somehow when saving the checkpoint. A complication with the `fold_activation_norm_scaling_factor()` approach is that we need to do it only on the version of the SAE weights being saved, not on the in-progress SAE being trained.

### Checklist

oli-clive-griffin:
# Description

Fixes #381

When training using `normalize_activations = "expected_average_only_in"`, ([Anthropic April Update](https://transformer-circuits.pub/2024/april-update/index.html#:~:text=The%20dataset%20is,different%20size%20transformers.)) we calculate a constant scaling factor which all subsequent batches are scaled by. The `ActivationsCache` then scales batches by this factor as they're loaded. This factor isn't a config variable, it's estimated at the start of training, so isn't persisted with the training config.

`SAETrainer` checkpoints during training, but because of the above, the calculated `estimated_norm_scaling_factor` isn't persisted, and so checkpoints are unusable.

The two options were:
1. fold the scaling factor into the weights before persisting. this way the persisted model is usable, however it's not obvious that they're able to resume being trained (at least not with the same hyperparameters).
2. persist `estimated_norm_scaling_factor`. This has the benefit of making it easier to resume training, (while this is not currently supported, this makes it possible to manually do) while also leaving the user able to explicitly fold the factor into the SAE weights if desired.

I chose to do the latter.

## Type of change

# Checklist:

- [ ] ~~I have made corresponding changes to the documentation~~
  - my added test produces a warning, but this is via an existing testing utility function. I can look into fixing this if that's an issue.


sae_lens/sae_training_runner.py
```diff
@@ -187,58 +185,56 @@ def _init_sae_group_b_decs(
...
-        config = trainer.cfg.to_dict()
-        with open(f"{path}/cfg.json", "w") as f:
+        config = self.cfg.to_dict()
+        with open(cfg_path, "w") as f:
             json.dump(config, f)
 
-        log_feature_sparsities = {"sparsity": trainer.log_feature_sparsity}
-
-        log_feature_sparsity_path = f"{path}/{SPARSITY_PATH}"
-        save_file(log_feature_sparsities, log_feature_sparsity_path)
```
oli-clive-griffin, line 214: this is handled by `self.sae.save_model`

sae_lens/training/sae_trainer.py
```diff
@@ -204,15 +204,6 @@ def fit(self) -> TrainingSAE:
...
 
-    @torch.no_grad()
-    def _estimate_norm_scaling_factor_if_needed(self) -> None:
-        if self.cfg.normalize_activations == "expected_average_only_in":
-            self.activation_store.estimated_norm_scaling_factor = (
-                self.activation_store.estimate_norm_scaling_factor()
-            )
-        else:
-            self.activation_store.estimated_norm_scaling_factor = 1.0
-
```
oli-clive-griffin, line 215: moved this into `ActivationsStore` as it's setting a property on it, from a method in it. Also having it in `ActivationsStore` keeps the handling of the lifecycle of `ActivationsStore.estimated_norm_scaling_factor` together

sae_lens/training/activations_store.py
```diff
@@ -234,7 +234,7 @@ def __init__(
 
         self.n_dataset_processed = 0
 
-        self.estimated_norm_scaling_factor = 1.0
+        self.estimated_norm_scaling_factor = None
```
oli-clive-griffin, line 237: Would love thoughts on this, I've changed this to initialise as `None` so that attempting to use it before explicitly setting it raises an error, which (as far as I can tell) would seem good as that would be an error case.

Unfortunately, `ActivationsStore.get_buffer` relies on `self.estimated_norm_scaling_factor` by way of `self.apply_norm_scaling_factor`. So we need it to have a value when loading the batches used to estimate the scaling factor (see comments below). My solution of temporarily setting and unsetting it is pretty ugly, so it might be best to simply keep this initialising as `1.0` and accept the aforementioned risk.

(I did some wider refactoring which got around this, but I decided not to submit them as they made this PR bigger and wider-reaching.)

sae_lens/training/activations_store.py
```diff
@@ -395,10 +395,22 @@ def load_cached_activation_dataset(self) -> Dataset | None:
...
+            raise ValueError(
+                "estimated_norm_scaling_factor is not set, call set_norm_scaling_factor_if_needed() first"
+            )
         return activations * self.estimated_norm_scaling_factor
 
     def unscale(self, activations: torch.Tensor) -> torch.Tensor:
+        if self.estimated_norm_scaling_factor is None:
+            raise ValueError(
+                "estimated_norm_scaling_factor is not set, call set_norm_scaling_factor_if_needed() first"
+            )
```
oli-clive-griffin, line 413: before, if calling either of these before setting `self.estimated_norm_scaling_factor`, it would work, which is _kind of_ a (user-caused) bug. this way that can't happen

sae_lens/training/activations_store.py
```diff
@@ -410,7 +422,10 @@ def estimate_norm_scaling_factor(self, n_batches_for_norm_estimate: int = int(1e
         for _ in tqdm(
             range(n_batches_for_norm_estimate), desc="Estimating norm scaling factor"
         ):
+            # temporalily set estimated_norm_scaling_factor to 1.0 so the dataloader works
+            self.estimated_norm_scaling_factor = 1.0
             acts = self.next_batch()
+            self.estimated_norm_scaling_factor = None
```
oli-clive-griffin, line 428: this is the aforementioned hack

sae_lens/training/activations_store.py
```diff
@@ -706,6 +721,10 @@ def state_dict(self) -> dict[str, torch.Tensor]:
         }
         if self._storage_buffer is not None:  # first time might be None
             result["storage_buffer"] = self._storage_buffer
+        if self.estimated_norm_scaling_factor is not None:
+            result["estimated_norm_scaling_factor"] = torch.tensor(
+                self.estimated_norm_scaling_factor
+            )
```
oli-clive-griffin, line 727: this is the central change

sae_lens/sae_training_runner.py
```diff
@@ -187,58 +185,56 @@ def _init_sae_group_b_decs(
...
-        path = f"{checkpoint_path}"
-        os.makedirs(path, exist_ok=True)
+        self.activations_store.save(str(base_path / "activation_store_state.json"))
 
+        # comment:
+        # Is this broken? it seems like this is "assymetrical" in the sense that
+        # this changes the output of the SAE.
+        # Should we not also balance this by scaling the encoder weights?
         if self.sae.cfg.normalize_sae_decoder:
             self.sae.set_decoder_norm_to_unit_norm()
```
oli-clive-griffin, line 201: I probably am missing something, it seems like this is "asymmetrical" in the sense that it changes the decoder without accordingly scaling the other SAE params. Should we not also balance this by scaling the other encoder weights?

chanind (commented):
Just a few minor comments, but otherwise refactorings look good! It's definitely an improvement to save the activation store state in the checkpoint, but seems like we'd need to also load that state in order to use the checkpoint. Given this is an improvement regardless, I'd say it's worth merging.

One issue though, on the final checkpoint save in `SAETrainer.fit()`, we first fold the norm into the SAE weights before saving. We should make sure to set the activations store `estimated_norm_scaling_factor` to 1.0 before saving the final checkpoint after doing that, since otherwise it will look like the final checkpoint requires re-folding the scaling factor when it's actually be done already.

sae_lens/sae_training_runner.py
oli-clive-griffin: ah interesting. A little off topic but could this optimizer instability be fixed by also scaling the other network parameters when the decoder norm is forced to 1? I guess there may be other optimizer instability introduced by scaling everything but should be better?

oli-clive-griffin: either way, are you suggesting we remove it here?

oli-clive-griffin:
> but seems like we'd need to also load that state in order to use the checkpoint

I was confused about this. It looks like this ability was removed here: https://github.com/jbloomAus/SAELens/pull/170/files#diff-0d417567c11e5e787c677c71e0500e1531d339964625fad3f48052076d25912eL509-L535, and resuming was removed here: https://github.com/jbloomAus/SAELens/pull/156.

So currently, the only training artefact really supported by the library is the final model. should we reintroduce resuming? or potentially remove checkpoints altogether? It seems we're in a weird middle ground. Alternatively, if we want to support loading an `ActivationsStore` outside the context of resuming, I can add the load function back in?

sae_lens/sae_training_runner.py
chanind: This has been fixed by normalizing the L1 penalty by the decoder norm (`cfg.scale_sparsity_penalty_by_decoder_norm`). This was from one of the Anthropic updates. This removes the pressure on the SAE to shrink the encoder / grow the decoder to reduce L1.

We should not delete this, since it's possible the decoder norm could be slightly larger than 1 without this here.

chanind:
re-introducing resuming is another discussion, but we shouldn't add that in this PR IMO. Personally, I'd like to be able to experiment with SAE checkpoints saved thoughout training, but can't do that currently due to this issue. Saving this scaling factor at least makes it possible to manually fix the checkpoints, but it is a bit annoying

oli-clive-griffin:
Sweet. In that case, looks like we're ready to merge?

chanind:
I think the final issue is the following:

> One issue though, on the final checkpoint save in SAETrainer.fit(), we first fold the norm into the SAE weights before saving. We should make sure to set the activations store estimated_norm_scaling_factor to 1.0 before saving the final checkpoint after doing that, since otherwise it will look like the final checkpoint requires re-folding the scaling factor when it's actually be done already.

Currently, the final checkpoint is saved with the norm folded in which is good, but this PR will also save the activation store state including the norm as if it hasn't been folded in. For the final checkpoint, we should make sure the norm is removed from the activation store before saving. This would ideally also have a test on it, saving 2 checkpoints, and assert that the norm is set on the first checkpoint but not the final checkpoint.

oli-clive-griffin:
Right, unless I'm misunderstanding you, this functionality is already implemented. However your test idea is a good one, I'll add that. 

https://github.com/jbloomAus/SAELens/pull/395/files#diff-a04ec095ce61475769ebc2c389af384a120a126ec0416566bd0f2be17816d2e0R194

chanind:
My bad, you're right! yeah if you don't mind adding the test this should be good to merge!

oli-clive-griffin:
sweet, will add that test now

PR #399, opened 2024-12-06
chore: fix tokenizer typing for bos_token_id

chanind:
# Description

It looks like `PreTrainedTokenizerBase` typing no longer has `bos_token_id` or other token IDs defined in the most recent version of transformers. I'm not sure if that was intentional by `transformers`, or an oversight, but this PR just ignore the typing for these token IDs so CI can pass.

## Type of change

- [x] CI chore

### DIFF
--- a/sae_lens/pretokenize_runner.py
+++ b/sae_lens/pretokenize_runner.py
@@ -60,11 +60,11 @@ def get_special_token_from_cfg(
     if isinstance(cfg_token, int):
         return cfg_token
     if cfg_token == "bos":
-        return tokenizer.bos_token_id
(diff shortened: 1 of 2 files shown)


chanind:
It looks like this was changed in https://github.com/huggingface/transformers/pull/34461/files#diff-85b29486a884f445b1014a26fecfb189141f2e6b09f4ae701ee758a754fddcc1, not sure if this is intentional

chanind:
issue opened here: https://github.com/huggingface/transformers/issues/35126

chanind:
merging as this should be pretty uncontroversial and CI is currently broken on main

PR #400, opened 2024-12-29
Replace assert statements with exception code

anthonyduong9:
# Description

Replaces `assert` statements with code to raise exceptions when I'd expect that a user could see an `AssertionError` and excluding `assert` statements for `pyright` and in notebooks/the `tutorials` directory. I'd made a comment in https://github.com/jbloomAus/SAELens/pull/226#discussion_r1835621811 about `assert` statements in production code and we want to change them to code to throw exceptions when we'd expect users could see them fail.

Fixes # (issue)

## Type of change

# Checklist:

### Performance Check.

- [ ] L0
- [ ] CE Loss
- [ ] MSE Loss


chanind (approved):
👍 Looks great!

PR #430, opened 2025-02-17
perf: support early-stopping in HF models

### LINKED ISSUE #429: [Proposal] Support early stopping in huggingface LLM wrapper
### Proposal 

We currently have a TODO in [load_model.py#L98](https://github.com/jbloomAus/SAELens/blob/main/sae_lens/load_model.py#L98) to support the `stop_at_layer` param in `HookedProxyLM` for huggingface models. Adding support for this will save compute when we're just trying to extract LLM activations at a specific layer, since there's no need to extract activations at later layers.

An example of how to do this is in [baukit/nethook.py](https://github.com/davidbau/baukit/blob/main/baukit/nethook.py#L189) . We just need to throw an exception after we process the layer we care about to stop processing, and then catch that exception before returning to the user.

### Checklist

chanind:
# Description

This PR supports early stopping in the huggingface wrapper. This is a perf improvement, so we don't need to run the whole model when all we want to do is extract some activations from an intermediate layer.

Fixes #429 

## Type of change

- [x] performance improvement

# Checklist:

### DIFF
--- a/sae_lens/load_model.py
+++ b/sae_lens/load_model.py
@@ -1,4 +1,7 @@
-from typing import Any, Literal, cast
+import re
+from collections import defaultdict
+from collections.abc import Iterable
+from typing import Any, Callable, Literal, cast
 
 import torch
 from transformer_lens import HookedTransformer
@@ -27,7 +30,7 @@ def load_model(
         if n_devices > 1:
             logger.info("MODEL LOADING:")
             logger.info("Setting model device to cuda for d_devices")
-            logger.info(f"Will use cuda:0 to cuda:{n_devices-1}")
+            logger.info(f"Will use cuda:0 to cuda:{n_devices - 1}")
             device = "cuda"
             logger.info("-------------")
 
@@ -72,6 +75,7 @@ class HookedProxyLM(HookedRootModule):
         super().__init__()
         self.model = model
         self.tokenizer = tokenizer
+        self.decoder_block_matcher = guess_decoder_block_matcher(model)
         self.setup()
 
     # copied and modified from base HookedRootModule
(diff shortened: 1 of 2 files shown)


sae_lens/load_model.py
```diff
@@ -95,20 +99,35 @@ def forward(
...
             raise NotImplementedError(
                 "Only return_type supported is 'both' or 'logits' to match what's in evals.py and ActivationsStore"
             )
-        output = self.model(tokens)
-        logits = _extract_logits_from_output(output)
+
+        stop_hook = None
+        if stop_at_layer is not None:
+            if return_type != "logits":
+                raise NotImplementedError(
```
rhaps0dy, line 114: The fact that it's a NotImplementedError instead of eg ValueError or RuntimeError implies that it can be supported in the future -- is that true?

If it's true what is it -- returning `(None, None)` instead?

chanind: Yeah there's nothing keeping this from being implemented in the future - it's just that SAELens currently doesn't use more of TLens API, so there's no point in implementing support for more options here beyond what SAELens actually uses. This class is just a wrapper around Huggingface models to make them look like a TLens model for SAELens to work wtih.

sae_lens/load_model.py
```diff
@@ -72,6 +75,7 @@ def __init__(self, model: torch.nn.Module, tokenizer: PreTrainedTokenizerBase):
         super().__init__()
         self.model = model
         self.tokenizer = tokenizer
+        self.decoder_block_matcher = guess_decoder_block_matcher(model)
```
anthonyduong9, line 78: Is effectively getting the layer names from `_guess_block_matcher_from_layers()` the best we can do? I'm afraid it's error-prone and implicit.

chanind: I'm not sure how else to know what the huggingface block name is for a given layer number otherwise since there's no standardized way to handle this. It seems like in practice nearly every LLM uses `model.layers.N` or `transformer.h.N` which this should handle without issue. The way we specify layer num in the config is a bit strange tbh, so maybe in the refactor we can fix that and just pass in a list of layer names we want to extract activations from rather than using the layer number. I'll think more about this...

chanind:
Done in v6.0.0

PR #432, opened 2025-02-22
fix: base k_aux on d_in instead of d_sae in topk aux loss

chanind:
# Description

This PR fixes a minor issue with our topk aux loss implementation, where we calculate `k_aux` using `d_sae` instead of the correct `d_in`. This likely doesn't make a huge difference in practice, but can't hurt to fix. This PR also calls `detach()` on the residual error before calculating topk aux loss, similar to [dictionary_learning's implementation](https://github.com/saprmarks/dictionary_learning/blob/main/dictionary_learning/trainers/top_k.py#L297). This should help ensure that the aux loss only pulls dead latents towards the SAE error, and doesn't accidentally pull live latents towards dead latents.

This PR also adds a test asserting that our topk aux loss matches the sparsity implementation aside from a normalization factor.

As a side note, we should probably add an optional `aux_coefficient` to the config to further customize this, but this is something that probably makes sense to hold off on until the refactor is done. 

## Type of change

# Checklist:

### DIFF
--- a/sae_lens/training/training_sae.py
+++ b/sae_lens/training/training_sae.py
@@ -473,7 +473,7 @@ class TrainingSAE(SAE):
             dead_neuron_mask is not None
             and (num_dead := int(dead_neuron_mask.sum())) > 0
         ):
-            residual = sae_in - sae_out
+            residual = (sae_in - sae_out).detach()
 
             # Heuristic from Appendix B.1 in the paper
             k_aux = sae_in.shape[-1] // 2
--- a/sae_lens/training/training_sae.py
+++ b/sae_lens/training/training_sae.py
@@ -469,31 +469,28 @@ class TrainingSAE(SAE):
     ) -> torch.Tensor:
         # Mostly taken from https://github.com/EleutherAI/sae/blob/main/sae/sae.py, except without variance normalization
         # NOTE: checking the number of dead neurons will force a GPU sync, so performance can likely be improved here
-        if (
-            dead_neuron_mask is not None
-            and (num_dead := int(dead_neuron_mask.sum())) > 0
-        ):
-            residual = (sae_in - sae_out).detach()
+        if dead_neuron_mask is None or (num_dead := int(dead_neuron_mask.sum())) == 0:
+            return sae_out.new_tensor(0.0)
+        residual = (sae_in - sae_out).detach()
 
-            # Heuristic from Appendix B.1 in the paper
-            k_aux = sae_in.shape[-1] // 2
+        # Heuristic from Appendix B.1 in the paper
+        k_aux = sae_in.shape[-1] // 2
 
-            # Reduce the scale of the loss if there are a small number of dead latents
-            scale = min(num_dead / k_aux, 1.0)
-            k_aux = min(k_aux, num_dead)
+        # Reduce the scale of the loss if there are a small number of dead latents
+        scale = min(num_dead / k_aux, 1.0)
+        k_aux = min(k_aux, num_dead)
 
-            auxk_acts = _calculate_topk_aux_acts(
-                k_aux=k_aux,
-                hidden_pre=hidden_pre,
-                dead_neuron_mask=dead_neuron_mask,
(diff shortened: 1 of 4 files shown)


sae_lens/training/training_sae.py
```diff
@@ -473,10 +473,10 @@ def calculate_topk_aux_loss(
             dead_neuron_mask is not None
             and (num_dead := int(dead_neuron_mask.sum())) > 0
         ):
-            residual = sae_in - sae_out
+            residual = (sae_in - sae_out).detach()
 
             # Heuristic from Appendix B.1 in the paper
-            k_aux = hidden_pre.shape[-1] // 2
+            k_aux = sae_in.shape[-1] // 2
```
anthonyduong9, line 479: Since we're changing `calculate_topk_aux_loss()`, want to also refactor it to use an early return?

It'd be

```
    def calculate_topk_aux_loss(
        self,
        sae_in: torch.Tensor,
        sae_out: torch.Tensor,
        hidden_pre: torch.Tensor,
        dead_neuron_mask: torch.Tensor | None,
    ) -> torch.Tensor:
        # Mostly taken from https://github.com/EleutherAI/sae/blob/main/sae/sae.py, except without variance normalization
        # NOTE: checking the number of dead neurons will force a GPU sync, so performance can likely be improved here
        if dead_neuron_mask is None or int(dead_neuron_mask.sum()) == 0:
            return sae_out.new_tensor(0.0)

        num_dead = int(dead_neuron_mask.sum())
        residual = (sae_in - sae_out).detach()

        # Heuristic from Appendix B.1 in the paper
        k_aux = sae_in.shape[-1] // 2

        # Reduce the scale of the loss if there are a small number of dead latents
        scale = min(num_dead / k_aux, 1.0)
        k_aux = min(k_aux, num_dead)

        auxk_acts = _calculate_topk_aux_acts(
            k_aux=k_aux,
            hidden_pre=hidden_pre,
            dead_neuron_mask=dead_neuron_mask,
        )

        # Encourage the top ~50% of dead latents to predict the residual of the
        # top k living latents
        recons = self.decode(auxk_acts)
        auxk_loss = (recons - residual).pow(2).sum(dim=-1).mean()
        return scale * auxk_loss
```

, which has a lot less indentation/is a lot easier to read.

tests/training/test_training_sae.py
```diff
@@ -101,28 +102,68 @@ def test_calculate_topk_aux_acts_k_less_than_dead():
...
 
 
-def test_TrainingSAE_calculate_topk_aux_loss():
-    # Create a small test SAE with d_sae=4, d_in=3
+def test_TrainingSAE_topk_aux_loss_matches_unnormalized_sparsify_implementation():
     cfg = build_sae_cfg(
-        d_in=3,
-        d_sae=4,
+        d_in=128,
+        d_sae=192,
```
anthonyduong9, line 108: Want to extract all the `128`s and `192`s to variables `d_in` and `d_sae`, respectively?

tests/training/test_training_sae.py
```diff
@@ -101,28 +102,68 @@ def test_calculate_topk_aux_acts_k_less_than_dead():
...
         architecture="topk",
+        activation_fn_kwargs={"k": 26},
         normalize_sae_decoder=False,
     )
 
+    sae = TrainingSAE(TrainingSAEConfig.from_sae_runner_config(cfg))
+    comparison_sae = SparseCoder(d_in=128, cfg=SparseCoderConfig(num_latents=192, k=26))
```
anthonyduong9, line 115: I think `sparse_coder` might be a better name than `comparison_sae` as someone reading the test could quickly tell it's an instance of `SparseCoder`.

anthonyduong9 (approved):

curt-tigges:
@chanind could you apply these changes to the new TopKTrainingSAE class in the refactor PR (once merged) as well?

chanind:
Will do! Once I manage to fix this strange triton error in tests...

PR #438, opened 2025-03-01
feat: Add support for dataset name in PretokenizeRunner

keltin13:
# Description

`PretokenizerRunner` doesn't support datasets `name` argument. Usually you can get around this by specifying the `data_dir`, but supporting this argument brings the interface more in line with the `datasets` interface. Additionally, the column name was previously not being tracked in the metadata, and now is. Finally, a pretokenizer test was using `pretokenize_runner` instead of `PretokenizerRunner`, which was deprecated, and has now been upgraded. 

Summary of updates:
- Added optional dataset name parameter
- Updated metadata to track dataset name and column name
- Added test for loading dataset with name argument and checking metadata
- Updated tests to use PretokenizeRunner instead of depricated pretokenizer_runner

Dependencies:
- Loads new (but small) dataset from HF for tests

Fixes # (issue)

## Type of change

# Checklist:

### DIFF
--- a/sae_lens/config.py
+++ b/sae_lens/config.py
@@ -674,6 +674,7 @@ def _validate_seqpos(seqpos: tuple[int | None, ...], context_size: int) -> None:
 class PretokenizeRunnerConfig:
     tokenizer_name: str = "gpt2"
     dataset_path: str = ""
+    dataset_name: str | None = None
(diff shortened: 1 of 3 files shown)


chanind:
Makes sense! Thanks for adding this and for the tests!

PR #443, opened 2025-03-10
Use conventional explained variance definition

Stefan-Heimersheim:
# Description

_Draft. I haven't tested these changes yet._

SAEBench & SAELens were using this explained variance / fraction of variance unexplained (FVU) formula:
image

But this is the more sensible / conventional formula (e.g. Wikipedia [here](https://en.wikipedia.org/wiki/Coefficient_of_determination) and [here](https://en.wikipedia.org/wiki/Fraction_of_variance_unexplained)):
image

For more discussion see [this LessWrong post](https://www.lesswrong.com/posts/ZBjhp6zwfE8o8yfni/stefanhex-s-shortform?commentId=95yixuEzraPh8mYTL).

I just fixed this in SAEBench ([PR here](https://github.com/adamkarvonen/SAEBench/pull/61)), and this is the same changes ported back to SAELens.

This PR changes the computation of the explained variance used in SAEBench, fixing two issues:
1. The nominator and denominator need to be averaged separately (this is what my LessWrong post was about)
2. The mean for the variance is averaged over all batches (I had omitted this in LessWrong post calculations, it's a smaller effect but fixed now)

FYI this is the formula I use for variance, to compute it in a batched way (this is just the multi-dimensional version of Var(x) = E[X^2] - E[X]^2, ChatGPT-derived, _k_ is the sample index, _a_ is the dimension index):
image

## Type of change

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [x] [Unsure] This change requires a documentation update

Renaming the current (unconventional / wrong) `explained_variance` to `explained_variance_legacy` is a breaking change (the value reported in `explained_variance` will change). We could not do this, and instead have an `explained_variance_new` or something but that's maybe worse?

# Checklist:

- [ ] New and existing tests pass locally with my changes


chanind:
Thank you for finding this and submitting this! I fixed the linting errors in the PR, added similar logic to the training loop (where variance explained is calculated per training batch), and added a simple test on a SAE that implements the identity function. I want to add more in-depth tests to evals but that will require refactoring the eval function to break it up into more easily testable chunks, and feels out of scope for this PR. Do the changes I made to your PR look correct? If so I'll merge this and then do the testing refactor in a follow-up PR.

stefan-apollo:
I've read through your changes and they look correct to me! Apologies for the delay (feel free to Slack/email me).

An initially-confusing point was that you removed `"explained_variance"` from the `metric_dict` -- I see what you did there now, maybe add a small comment explaining why explained variance is handled separately.

chanind:
I'll merge this once we resolve the issue with CI failing, which seems like it's caused by huggingface adding a new rate limit on loading datasets that tests are violating.

PR #393, opened 2024-11-30
Replace isort black and flake8 with ruff

anthonyduong9:
# Description

Replaces `isort`, `black`, and `flake8` with Ruff, and adds more lint rules. Ruff is a lot faster, replaces multiple dependencies with one, and supports lint rules. I'd made a comment in https://github.com/jbloomAus/SAELens/pull/226#discussion_r1835620125 about style and found out we weren't using `.pylintrc`, and so we want to add lint rules to enforce style.

Fixes #368 

## Type of change

Please delete options that are not relevant.

- [x] Bug fix (non-breaking change which fixes an issue)

# Checklist:

- [ ] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes
- [x] I have not rewritten tests relating to key interfaces which would affect backward compatibility

### You have tested formatting, typing and unit tests (acceptance tests not currently in use)

- [x] I have run `make check-ci` to check format and linting. (you can run `make format` to format code if needed.)

### Performance Check.

If you have implemented a training change, please indicate precisely how performance changes with respect to the following metrics:
- [ ] L0
- [ ] CE Loss
- [ ] MSE Loss
- [ ] Feature Dashboard Interpretability

Please links to wandb dashboards with a control and test group.

Issue #311, opened 2024-10-01
[Bug Report] Unable to train Mamba SAE


joelburget:
If you are submitting a bug report, please fill in the following details and use the tag [bug].

**Describe the bug**
**Code example**
```
cfg = LanguageModelSAERunnerConfig(
    model_name="state-spaces/mamba-2.8b",
    model_class_name="HookedMamba",
    ...
)
sae = SAETrainingRunner(cfg).run()
```

```
Traceback (most recent call last):
  File "/workspace/mamba-sae/hyperparam_sweep.py", line 70, in <module>
    sae = SAETrainingRunner(cfg).run()
  File "/usr/local/lib/python3.10/dist-packages/sae_lens/sae_training_runner.py", line 82, in __init__
    self._init_sae_group_b_decs()
  File "/usr/local/lib/python3.10/dist-packages/sae_lens/sae_training_runner.py", line 170, in _init_sae_group_b_decs
    layer_acts = self.activations_store.storage_buffer.detach()[:, 0, :]
  File "/usr/local/lib/python3.10/dist-packages/sae_lens/training/activations_store.py", line 376, in storage_buffer
    self._storage_buffer = self.get_buffer(self.half_buffer_size)
  File "/usr/local/lib/python3.10/dist-packages/torch/utils/_contextlib.py", line 115, in decorate_context
    return func(*args, **kwargs)
  File "/usr/local/lib/python3.10/dist-packages/sae_lens/training/activations_store.py", line 545, in get_buffer
    refill_batch_tokens = self.get_batch_tokens(
  File "/usr/local/lib/python3.10/dist-packages/sae_lens/training/activations_store.py", line 410, in get_batch_tokens
    return torch.stack(sequences, dim=0).to(self.model.W_E.device)
  File "/usr/local/lib/python3.10/dist-packages/torch/nn/modules/module.py", line 1688, in __getattr__
    raise AttributeError(f"'{type(self).__name__}' object has no attribute '{name}'")
AttributeError: 'HookedMamba' object has no attribute 'W_E'. Did you mean: 'W_K'?
```

<details>
<summary>Full output</summary>

```
root@39393c1a11a5:/workspace/mamba-sae# wandb agent PEAR-ML/mamba-sae-sweep/i0mhj40q
wandb: Starting wandb agent 🕵️
2024-10-01 15:28:00,877 - wandb.wandb_agent - INFO - Running runs: []
[...]

PR #449, opened 2025-03-20
removes .flake8

anthonyduong9:
# Description

Removes `.flake8`. This is unused after https://github.com/jbloomAus/SAELens/pull/393.

Fixes # (issue)

## Type of change

# Checklist:

### Performance Check.

- [ ] L0
- [ ] CE Loss
- [ ] MSE Loss


requirements.txt
```diff
@@ -7,9 +7,6 @@ nbformat==5.9.2
 ipykernel==6.27.1
 matplotlib==3.8.2
 matplotlib-inline==0.1.6
-flake8==7.0.0
-isort==5.13.2
-black==23.11.0
```
chanind, line 12: Do we even need requirements.txt at all?

chanind (approved):
👍

requirements.txt
anthonyduong9: I don't think so. Though it looks like one person installed it, in https://github.com/jbloomAus/SAELens/issues/311#issue-2559590681. I think it's best I delete it though, since it has clearly become out-of-sync with `pyproject.toml`, i.e., wrong.

PR #467, opened 2025-04-26
fix: Use the whole batch when estimating norm scaling factor.

mkbehr:
# Description

`ActivationsStore.estimate_norm_scaling_factor()` was taking the first sample from each batch, when it looks like it was meant to take the first (and only) layer. This fixes it to use the whole batch instead.

## Type of change

# Checklist:

### Performance Check.

- [ ] L0
- [ ] CE Loss
- [ ] MSE Loss


chanind (approved):
Great catch! Thank you for this fix

PR #474, opened 2025-05-05
fix: heuristic init should place latents anywhere in the space

chanind:
# Description

This PR fixes a bug with the `decoder_heuristic_init` option. The decoder should be initialized to random directions (see: https://transformer-circuits.pub/2024/april-update/index.html#training-saes). However, SAELens was using `torch.rand`, which picks values at random between 0 and 1, rather than `torch.randn` which positive or negative values according to a normal distribution. As a result, the decoder had only positive values when using `decoder_heuristic_init`, which basically forces all the decoder latents into a tiny portion of the full subspace of size `1/d_in` of the full `d_in` vector space.

This PR fixes this issue and adds a test so we can't accidentally revert back to this behavior.

## Type of change

# Checklist:

### DIFF
--- a/tests/training/test_sae_initialization.py
+++ b/tests/training/test_sae_initialization.py
@@ -177,11 +177,11 @@ def test_SparseAutoencoder_initialization_heuristic_init():
 
     sae = TrainingSAE.from_dict(cfg.get_training_sae_cfg_dict())
 
-    decoder_norms = sae.W_dec.norm(dim=1)
-
     # Check for both positive and negative values in the decoder weights
     assert torch.any(sae.W_dec > 0)
     assert torch.any(sae.W_dec < 0)
+
+    decoder_norms = sae.W_dec.norm(dim=1)
 
     # not unit norms
     assert not torch.allclose(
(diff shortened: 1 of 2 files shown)


tests/training/test_sae_initialization.py
```diff
@@ -179,6 +179,10 @@ def test_SparseAutoencoder_initialization_heuristic_init():
 
     decoder_norms = sae.W_dec.norm(dim=1)
 
+    # Check for both positive and negative values in the decoder weights
+    assert torch.any(sae.W_dec > 0)
+    assert torch.any(sae.W_dec < 0)
```
anthonyduong9, line 184:
```suggestion
    # Check for both positive and negative values in the decoder weights
    assert torch.any(sae.W_dec > 0)
    assert torch.any(sae.W_dec < 0)

    decoder_norms = sae.W_dec.norm(dim=1)
```

I find it more readable to have the `assert` statements with `sae.W_dec` closer to the declaration of `sae` and the declaration of `decoder_norms` closer to the `assert` statements for them.

anthonyduong9 (approved):

PR #433, opened 2025-02-23
feat: custom loading converters for SAE.from_pretrained() and SAE.load_from_disk()

chanind:
# Description

This PR standardizes the interface for `PretrainedSaeHuggingfaceLoader` and `PretrainedSaeDiskLoader` to allow directly loading any custom SAE from another library as long as a correct `converter` param is passed to `SAE.from_pretrained()` or  `SAE.load_from_disk()`. 

This PR also deprecates the `SAE.load_from_pretrained()` method in favor of `SAE.load_from_disk()` to make the difference between `from_pretrained()`, which loads from Huggingface, and `load_from_disk()`, which loads a locally saved SAE more obvious.

These follow the following protocols:

```
class PretrainedSaeHuggingfaceLoader(Protocol):
    def __call__(
        self,
        repo_id: str,
        folder_name: str,
        device: str,
        force_download: bool,
        cfg_overrides: dict[str, Any] | None,
    ) -> tuple[dict[str, Any], dict[str, torch.Tensor], Optional[torch.Tensor]]: ...


class PretrainedSaeDiskLoader(Protocol):
    def __call__(
        self,
        path: str | Path,
        device: str,
        cfg_overrides: dict[str, Any] | None,
    ) -> tuple[dict[str, Any], dict[str, torch.Tensor]]: ...
```

Essentially, the provided function must take in either a `repo_id` and `folder_name` on huggingface, or a local path, optional config overrides, and return a config dict and model state dict.

Our current converters don't feel general enough to work for all instances of other libraries, but we can and should write some converters for the `dictionary_learning` and `sparsify` libraries in the future.

## Type of change

Please delete options that are not relevant.

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

# Checklist:

- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing tests pass locally with my changes
- [x] I have not rewritten tests relating to key interfaces which would affect backward compatibility

### You have tested formatting, typing and tests

- [x] I have run `make check-ci` to check format and linting. (you can run `make format` to format code if needed.)

PR #476, opened 2025-05-09
adds sparsify_huggingface_loader()

anthonyduong9:
# Description

Adds functions to load SAEs trained with the `sparsify` library, from Hugging Face or disk. In #433, we made a change so that you can load an SAE trained using any library as long as you pass a correct arg to the `converter` param of `SAE.from_pretrained()`/`SAE.load_from_disk()`, and said we wanted to add this for SAEs trained with the `sparsify` library.

Fixes # (issue)

## Type of change

# Checklist:

### Performance Check.

- [ ] L0
- [ ] CE Loss
- [ ] MSE Loss

### DIFF
--- a/sae_lens/config.py
+++ b/sae_lens/config.py
@@ -33,6 +33,7 @@ HfDataset = DatasetDict | Dataset | IterableDatasetDict | IterableDataset
 
 SPARSITY_FILENAME = "sparsity.safetensors"
 SAE_WEIGHTS_FILENAME = "sae_weights.safetensors"
+SPARSIFY_WEIGHTS_FILENAME = "sae.safetensors"
 SAE_CFG_FILENAME = "cfg.json"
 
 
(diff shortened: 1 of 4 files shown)


chanind (commented):
We actually have sparsity as a dev depenendency to we can assert stuff against the sparsify implementation. It should be pretty easy to add a test where you create sparsify SAE, save it to disk, then load it with the sparsify loader and verify it matches the original SAE.

anthonyduong9:
> We actually have sparsify as a dev depenendency to we can assert stuff against the sparsify implementation. It should be pretty easy to add a test where you create sparsify SAE, save it to disk, then load it with the sparsify loader and verify it matches the original SAE.

Good idea! I've added a test for `sparsify_huggingface_loader()` that asserts its output matches what you'd get from loading from Hugging Face using the `sparsify` library, as well as the test you described.

chanind (approved):
LGTM!

PR #483, opened 2025-05-19
Fix abnormally large llamascope L0.

Hzfinfdu:
# Description

We have received bug reports that Llamascope SAEs are not loaded properly, with large L0's on all tests. We have not identified which commit led to this incompatibility. Now with jumprelu threshold scaled with norm_scaling_factor they should work fine.

Fixes # (issue)

## Type of change

# Checklist:

### Performance Check.

- [ ] L0
- [ ] CE Loss
- [ ] MSE Loss

### DIFF
--- a/sae_lens/toolkit/pretrained_sae_loaders.py
+++ b/sae_lens/toolkit/pretrained_sae_loaders.py
@@ -476,11 +476,14 @@ def get_llama_scope_config_from_hf(
     # Get norm scaling factor to rescale jumprelu threshold.
     # We need this because sae.fold_activation_norm_scaling_factor folds scaling norm into W_enc.
     # This requires jumprelu threshold to be scaled in the same way
-    norm_scaling_factor = d_in ** .5 / old_cfg_dict['dataset_average_activation_norm']['in']
-    
+    norm_scaling_factor = (
+        d_in**0.5 / old_cfg_dict["dataset_average_activation_norm"]["in"]
+    )
+
     cfg_dict = {
         "architecture": "jumprelu",
-        "jump_relu_threshold": old_cfg_dict["jump_relu_threshold"] * norm_scaling_factor,
+        "jump_relu_threshold": old_cfg_dict["jump_relu_threshold"]
+        * norm_scaling_factor,
         # We use a scalar jump_relu_threshold for all features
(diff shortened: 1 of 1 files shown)


chanind:
I'm a bit confused what could have gone wrong here. Are you saying the SAEs work correctly before calling `fold_activation_norm_scaling_factor()` but after calling that, then things break? If so, this sounds like there's a bug with `fold_activation_norm_scaling_factor()` rather than something specific to Llamascope. `fold_activation_norm_scaling_factor()` shouldn't affect `b_enc` or the jumprelu threshold though, and we do have tests that should cover JumpReLU at https://github.com/jbloomAus/SAELens/blob/75d040f63bd86cf50637b6ea4f7fe667116b81d8/tests/training/test_sae_basic.py#L192.

Hzfinfdu:
Sorry for the late response. I am 99% certain that this won't introduce bugs in SAELens since this PR only changes Llamascope loading. But I am not certain which commit changed llamascope L0. Intuitively this may result from that llamascope SAEs were not trained with standard JumpReLU. Their JL threshold is determined post-training and hard-coded in the config. And `fold_activation_norm_scaling_factor()` or somewhere else changed, leading to such incompatibility.

chanind:
I believe that this won't break anything, but my fear was that we had introduced a bug in SAELens somehow, and if so, , we should just fix the source bug rather than putting a band-aid on the problem in the llamascope loader. Plus, if it was a bug in SAELens, then if we fix the bug it will then break llamascope again, especially since there's no tests on llamascope stuff. I tested in [the Colab notebook](https://colab.research.google.com/drive/1cCiRJbUbEibw8GXygJLaFWbioK_jXxwh?usp=sharing) what the llamascope behavior was in v4.3.0 which is the first time llamascope was released, and compared the L0 to now, and found that there has been no change in SAELens behavior. This makes me more confident that SAELens did not have a bug and your fix to llamascope isn't going to paper-over any deeper problems.

chanind:
Just to clarify - if you do find that something *has* changed, then we should fix the underlying source issue, not mess with llamascope loaders. Also if you can find a way to test that llamascope is loading correct in a unit test that's cheap enough to run in CI, that would also help guarantee that llamascope cannot break in the future.

Hzfinfdu:
I see. Thanks!

PR #492, opened 2025-06-10
adds fields for sequence logic in ActivationsStore

### LINKED ISSUE #472: [Proposal] ActivationStore flags to deal with BOS
### Proposal 

A flag for ActivationStore in SAELens to not add BOS between concatenated examples/prompts.
Also, a flag for ActivationStore in SAELens to never concatenate sequences and ignore examples that are shorter than requested.

### Motivation

From SAEDashboard:

Background
- Gemma 2 was not trained with BOS tokens

Problem
- When we use ActivationStore to generate prompts, often it concatenates multiple prompts together, resulting in <bos> at middle of text
- When some latents in Gemma see this, it gets confused and activates highly on the <bos>
- We end up with incorrect dashboards because top activations end up being <bos>, which mask the "true" activations

Context: https://github.com/jbloomAus/SAEDashboard/issues/46

### Checklist

anthonyduong9:
# Description

Adds a field, where if the value is `True`, excludes the BOS token between concatenated sequences, and a field, where if the value is `True`, disables concatenating sequences and ignores sequences shorter than the context size. Gemma 2 wasn't trained on BOS tokens and we want to match this during training. We can do this using `PretokenizeRunner`, but not without it.

Fixes #472

## Type of change

# Checklist:

### Performance Check.

- [ ] L0
- [ ] CE Loss
- [ ] MSE Loss

### DIFF
--- a/sae_lens/config.py
+++ b/sae_lens/config.py
@@ -150,6 +150,8 @@ class LanguageModelSAERunnerConfig(Generic[T_TRAINING_SAE_CONFIG]):
         sae_lens_version (str): The version of the sae_lens library.
         sae_lens_training_version (str): The version of the sae_lens training library.
         exclude_special_tokens (bool | list[int]): Whether to exclude special tokens from the activations. If True, excludes all special tokens. If a list of ints, excludes those token IDs.
+        disable_concat_sequences (bool): Whether to disable concatenating sequences and ignore sequences shorter than the context size. If True, disables concatenating and ignores short sequences.
+        exclude_bos_between_sequences (bool): Whether to exclude the BOS token between concatenated sequences. If True, excludes the BOS token.
     """
 
     sae: T_TRAINING_SAE_CONFIG
--- a/sae_lens/config.py
+++ b/sae_lens/config.py
@@ -46,6 +46,29 @@ def dict_field(default: dict[str, Any] | None, **kwargs: Any) -> Any:  # type: i
     return simple_parsing.helpers.dict_field(default, type=json_dict, **kwargs)
 
 
+def special_token(s: str) -> Any:
+    """Parse special token value from string."""
+    if s.lower() == "none":
+        return None
+    if s in ["bos", "eos", "sep"]:
+        return s
+    try:
+        return int(s)
+    except ValueError:
+        raise ValueError(
+            f"Expected 'bos', 'eos', 'sep', an integer, or 'none', got {s}"
+        )
+
+
+def special_token_field(
+    default: int | Literal["bos", "eos", "sep"] | None, **kwargs: Any
+) -> Any:  # type: ignore
+    """
+    Helper to wrap simple_parsing.helpers.field so we can load special token fields from the command line.
+    """
+    return simple_parsing.helpers.field(default=default, type=special_token, **kwargs)
+
+
 @dataclass
 class LoggingConfig:
     # WANDB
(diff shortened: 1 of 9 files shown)


sae_lens/training/activations_store.py
```diff
@@ -374,15 +380,23 @@ def _iterate_tokenized_sequences(self) -> Generator[torch.Tensor, None, None]:
...
+            if self.disable_concat_sequences:
+                for tokens in self._iterate_raw_dataset_tokens():
+                    if len(tokens) >= self.context_size:
+                        yield tokens[: self.context_size]
+            else:
+                sequence_separator_token_id = None
+                if self.prepend_bos and not self.exclude_bos_between_sequences:
+                    sequence_separator_token_id = bos_token_id
+
+                yield from concat_and_batch_sequences(
```
chanind, line 393: It would be better to adapt the `concat_and_batch_sequences` function to handle `disable_concat_sequences`, so then the pretokenize runner could also gain this capability too.

sae_lens/config.py
```diff
@@ -179,6 +181,8 @@ class LanguageModelSAERunnerConfig(Generic[T_TRAINING_SAE_CONFIG]):
     training_tokens: int = 2_000_000
     store_batch_size_prompts: int = 32
     seqpos_slice: tuple[int | None, ...] = (None,)
+    disable_concat_sequences: bool = False
+    exclude_bos_between_sequences: bool = False
```
chanind, line 185: Does it make sense to have the same options both here and in `pretokenize_runner`? I could see an argument that the `pretokenize_runner` arguments are too confusing / complicated if it's best to stick with just `bos` here, but can also see it being nice if we have the same options and thus same capabilities regardless of pretokenizing or not

chanind: Yeah I was thinking the latter, it is sort of strange these have different options, but can see the argument that maybe the pretokenize runner version is too complex if we want to keep them separate. It looks like the 6.0 release will have a lot of breaking changes so if we're going to unify these now is probably a good time to do it.

chanind (approved):
LGTM!

anthonyduong9:
@chanind I realized I didn't change the logic for `ActivationsStore.from_sae()` and `ActivationsStore.from_cache_activations()`.

I'm thinking I should add `disable_concat_sequences` and `sequence_separator_token` to `SAEMetadata`, and then change `ActivationsStore.from_sae()` to read those fields and pass the values to `cls()`, so that the logic for concatenating and ignoring sequences, and the sequence separator token, when loading a given SAE, are the same as they were while the SAE was trained.

And then for `ActivationsStore.from_cache_activations()`, I'm thinking we don't need to change anything.

Does that sound good to you?

chanind:
Sounds reasonable!

anthonyduong9:
Because of https://github.com/jbloomAus/SAEDashboard/issues/46#issuecomment-3021269415, I've actually made changes so `ActivationsStore.from_sae()` has params `disable_concat_sequences` and `sequence_separator_token`. For SAEs not trained with SAELens, callers need to be able to pass values without reading from `SAEMetadata`. Because of this, and that it's tricky for the function to distinguish between `None` passed by the caller and no value passed at all, it's tricky to make the `SAEMetadata` values the default. For SAEs trained with SAELens, the caller will have to pass the values that were passed for training (which are on `SAEMetadata`), if they're to be the same for loading.

chanind:
You should be able to check if a field is explicitly set on `SAEMetadata` with the `in` operator, e.g.: https://github.com/jbloomAus/SAELens/blob/main/tests/saes/test_sae.py#L72-L79. But regardless, letting the user pass in options seems fine. IIRC `ActivationsStore.from_sae()` is only ever used if the user is running evals on a random SAE from CLI anyway.

anthonyduong9:
In the case of `sequence_separator_token`, what we have now (a default value of `"bos"` for backwards compatibility, no token when `None` is passed, and otherwise, whatever value the caller passes) is easiest to reason about, IMO, so I'll just merge this as-is.

PR #499, opened 2025-07-13
WIP: Final 6.0.0 changes

chanind:
# Description

This PR fixes any remaining issues with the 6.x.x release before we do a final release.

This is currently WIP, but includes the following changes:

- Changes `SAE.from_pretrained()` to return the SAE directly instead of a tuple. The previous functionality is now called `SAE.from_pretrained_with_cfg_and_sparsity()`.
- changes the default adam_beta1 param to be 0.9
- fixes outstanding places where the sae.metadata was not being read correctly
- adds more tests to cover places with sae.metadata was not being read correctly
- updates some of the notebook with fixes (still more notebooks to go through)

### DIFF
--- a/benchmark/test_eval_all_loadable_saes.py
+++ b/benchmark/test_eval_all_loadable_saes.py
@@ -60,7 +60,7 @@ def test_loading_pretrained_saes(
     else:
         device = "cpu"
 
-    sae, _, _ = SAE.from_pretrained(release, sae_name, device=device)
+    sae = SAE.from_pretrained(release, sae_name, device=device)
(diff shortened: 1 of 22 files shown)


benchmark/test_eval_all_loadable_saes.py
anthonyduong9, line 1: I think we need to change `sae.cfg.hook_name` and `sae.cfg.model_name` to `sae.cfg.metadata.hook_name` and `sae.cfg.metadata.model_name`.

chanind:
aah I guess we still need the `layer_norm` normalization option, I'll add that back

PR #502, opened 2025-07-14
chore: updating docs for v6 release

chanind:
# Description

This PR makes minor changes to the docs around the v6 release:

- removes the `--pre` option from `pip install sae-lens`
- adds a note that `SAE.from_pretrained()` will continue to work with a deprecation warning
- changes the docs banner to point to the mitgration guide instead of the pre-release page


chanind:
merging as this should be pretty uncontroversial

PR #505, opened 2025-07-15
feat: BatchTopK SAE training

chanind:
# Description

This PR adds support for [BatchTopK](https://arxiv.org/abs/2412.06410) SAE training. These SAEs save as JumpReLU SAEs in inference mode.

## Type of change

# Checklist:

### DIFF
--- a/tests/helpers.py
+++ b/tests/helpers.py
@@ -4,6 +4,7 @@ from typing import Any, Sequence, TypedDict, cast
 from transformer_lens import HookedTransformer
 
 from sae_lens.config import LanguageModelSAERunnerConfig, LoggingConfig
+from sae_lens.saes.batchtopk_sae import BatchTopKTrainingSAEConfig
 from sae_lens.saes.gated_sae import GatedSAEConfig, GatedTrainingSAEConfig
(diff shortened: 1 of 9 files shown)


sae_lens/saes/topk_sae.py
anthonyduong9, line 1: We should also change the line

    then optionally applies a post-activation function (e.g., ReLU).

to

    then applies ReLU.
.

Also in this file, we can update all comments so they don't reference "noise", right?

tests/helpers.py
anthonyduong9, line 1: We could also update

- `ALL_ARCHITECTURES`
- `build_sae_cfg_for_arch()`
- `build_sae_training_cfg_for_arch()`
- `build_runner_cfg_for_arch()`

for `"batchtopk"`, which'll add test cases for it, if you think it's worth it.

In any case, the variable name `ALL_ARCHITECTURES` seems misleading if `"batchtopk"` isn't in it.

anthonyduong9 (approved):
LGTM once feedback is addressed.

sae_lens/saes/topk_sae.py
chanind: Indeed! I'll update this

PR #509, opened 2025-07-24
Feature/basic transcoders

curt-tigges:
# Description

Provides transcoder loading support for basic transcoders.

Fixes # (issue)

## Type of change

# Checklist:


chanind (commented):
Adding support for training transcoders requires a bunch more changes throughout that codebase that this PR doesn't seem to address. This requires that the trainer be able to handle input and output activations coming from different points in the model,  the activations store being able to supply those cross-layer activations, and the evals being aware of this as well. We should probably add a `llm_transcoder_training_runner` to handle transcoder-specific training options. We also will need tests to test that training works on Transcoders. The current tests in this PR only address running transcoders. IMO we should add support for loading / inference with transcoders  in its own PR before we add support for training, as training is a much bigger change.

hijohnnylin:
> The current tests in this PR only address running transcoders. IMO we should add support for loading / inference with transcoders in its own PR before we add support for training, as training is a much bigger change.

Agreed. Breaking the transcoders work down to loading + inference vs training is ideal. Does this complete the first part of loading + inference or is there any remaining work left that?

chanind:
It looks reasonable for loading / inference, but the majority of the PR is training code that is only partially implemented and will not work if run. IMO we should remove that code from this PR

hijohnnylin:
That sounds reasonable to me! @curt-tigges can you please move the training code to a separate branch + draft PR that can be picked up later by someone else? This sounds like it will unblock merging this PR. Thanks!

curt-tigges:
Training stubs removed.

sae_lens/saes/transcoder.py
```diff
@@ -0,0 +1,383 @@
...
+        feature_acts = self.encode(x)
+        sae_out = self.decode(feature_acts)
+        return sae_out, feature_acts
+
+    @property
+    def d_out(self) -> int:
+        """Output dimension of the transcoder."""
+        return self.cfg.d_out
+
+    def get_output_hook_name(self) -> str:
```
chanind, line 184: nit: we should remove this, this is TLens-specific

tests/saes/test_transcoder.py
```diff
@@ -0,0 +1,352 @@
...
+        assert transcoder_moved.W_dec.device == torch.device("meta")
+        assert transcoder_moved.b_enc.device == torch.device("meta")
+        assert transcoder_moved.b_dec.device == torch.device("meta")
+
+    def test_transcoder_fold_w_dec_norm(self):
+        cfg = build_transcoder_cfg()
+        transcoder = Transcoder(cfg)
+
+        # Set weights to non-unit norm
+        transcoder.W_dec.data = torch.randn_like(transcoder.W_dec) * 2.0
```
chanind, line 217: We should ensure every parameter is set randomly, if anything is zeros this test might pass while silently breaking the underlying transcoder. See `test_sae_fold_w_dec_norm_all_architectures` for an example of how to test this while initializing all parameters randomly.

Alternatively, maybe it's better to add `transcoder`, `jumprelu_transcoder`, and `skip_transcoder` to the `ALL_ARCHITECTURES` list in tests so these automatically get tested with the rest of the SAE classes. Or maybe we could make a new `ALL_ARCHITECTURES_INCLUDING_TRANSCODERS` list or something for tests that apply to everything like this.

tests/saes/test_transcoder.py
```diff
@@ -0,0 +1,352 @@
...
+    """Integration tests for transcoders."""
+
+    def test_transcoder_different_dimensions(self):
+        """Test transcoder with various input/output dimension combinations."""
+        test_cases = [
+            (32, 64, 48),  # d_in < d_sae, d_out between
+            (64, 32, 96),  # d_in > d_sae, d_out > d_in
+            (48, 96, 48),  # d_in == d_out
+            (128, 256, 64),  # all different, d_out < d_in
+        ]
```
chanind, line 313: nit: use `@pytest.mark.parametrize` to make parametrized tests like this

PR #525, opened 2025-08-19
move automated-interpretability to an optional dependency in sae-lens

### LINKED ISSUE #517: [Proposal] Update dependencies so Colab does not need to reinstall torch when installing SAELens
### Proposal 

Currently, in Colab, installing SAELens `!pip install sae-lens` will re-install torch and a number of other dependencies, then requests the user restart colab. This should not be necessary, and is likely due to us having some dependency have too strict of a bounds. We should figure out what the dependency is and fix the bounds to ease the use of SAELens in Colab.

chanind:
# Description

The `automated-interpretability` dependency causes a bunch of issues since it has very strict dependency requirement that are not frequently updated, and this means that SAELens thus can't be used with a lot of model python packages (e.g. see https://github.com/hijohnnylin/automated-interpretability/pull/4). `automated-interpretability` is only needed for Neuronpedia, so there's no reason normal users of SAELens should need this dependency anyway. This PR just moves the dependency to be optional, and can be installed with `pip install sae-lens[neuronpedia]`.

Fixes #517

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -28,12 +28,12 @@ babe = "^0.0.7"
 nltk = "^3.8.1"
 safetensors = ">=0.4.2,<1.0.0"
 mamba-lens = { version = "^0.0.4", optional = true }
-automated-interpretability = ">=0.0.5,<1.0.0"
 python-dotenv = ">=1.0.1"
 pyyaml = "^6.0.1"
 typing-extensions = "^4.10.0"
 simple-parsing = "^0.1.6"
 tenacity = ">=9.0.0"
+automated-interpretability = { version = ">=0.0.5,<1.0.0", optional = true }
 
 [tool.poetry.group.dev.dependencies]
 pytest = "^8.0.2"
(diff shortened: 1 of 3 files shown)


hijohnnylin:
Approved, but tbh I don't even think this package is used? (just searched for usages and couldnt find any though I could have missed something) I'm totally fine with removing it completely.

chanind:
its imports are called `neuron_explainer`, they're all in `neuronpedia_integration.py`

hijohnnylin:
Ahh I see. Yeah those are super old and I should remove at some point (ppl should just be using the pypi neuronpedia library instead), but ok to make this optional for now.

hijohnnylin:
Feel free to merge this whenever. Looks like some type checking issue, lmk if you want me to help resolve, happy to do so.

hijohnnylin (approved):
Looks good pending the type-checking stuff.

chanind:
Ah yeah, if there's an updated way to interact with neuronpedia probably good to switch to that at some point. Maybe can make an issue for it

PR #535, opened 2025-08-29
fix: add config loader for mntss clt by layer

hijohnnylin:
# Description

My previous commit https://github.com/jbloomAus/SAELens/pull/534 broke deploy docs because the doc build assumes that each pretrained_sae_loader has a corresponding config_getter.
This makes the config_getter for it, which basically just calls the sae_loader and gets the first value in the tuple.

Since the mntss CLT loader requires loading weights (and config getting is presumed to be a "lightweight" operation"), I've also added a warning log.

Fingers crossed that this doesn't OOM or timeout the CI!

## Type of change

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -32,10 +32,10 @@ jobs:
       - name: Cache Huggingface assets
         uses: actions/cache@v4
         with:
-          key: huggingface-3-${{ runner.os }}-${{ matrix.python-version }}-${{ hashFiles('**/pyproject.toml') }}
+          key: huggingface-4-${{ runner.os }}-${{ matrix.python-version }}-${{ hashFiles('**/pyproject.toml') }}
           path: ~/.cache/huggingface
           restore-keys: |
-            huggingface-3-${{ runner.os }}-${{ matrix.python-version }}-
+            huggingface-4-${{ runner.os }}-${{ matrix.python-version }}-
       - name: Load cached Poetry installation
         id: cached-poetry
         uses: actions/cache@v4
--- a/sae_lens/loading/pretrained_sae_loaders.py
+++ b/sae_lens/loading/pretrained_sae_loaders.py
@@ -1393,7 +1393,7 @@ def mntss_clt_layer_huggingface_loader(
 def get_mntss_clt_layer_config_from_hf(
     repo_id: str,
     folder_name: str,
-    device: str | None = None,
+    device: str,
     force_download: bool = False,  # noqa: ARG001
     cfg_overrides: dict[str, Any] | None = None,
 ) -> dict[str, Any]:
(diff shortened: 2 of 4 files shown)


hijohnnylin:
Oof this indeed makes us run out of space on the runner :( 
https://github.com/jbloomAus/SAELens/actions/runs/17334324457/job/49217362021?pr=535

I will look into a better solution soon

chanind:
Does this need to be in the `pretrained_saes.yaml`? It's sort of a weird pseudo-transcoder that only really makes sense for generating dashboards since the decoders are all collapsed together. This is probably not super useful to the general public outside of this use-case and shouldn't be used as a stand-alone transcoder.

hijohnnylin:
Hmmm that's a good point. I should have checked with you before merging this, sorry!

Our use case for this is that we want to allow activation testing with custom prompts for these on Neuronpedia (and to load them into Neuronpedia in general). Eg people can see what's on the dashboard, but also be able to enter their own prompts to verify. The simplest way for us to access this was to just treat it like all the other SAEs on SAELens. But you are right that this could be not a good fit for the way SAELens is intended to be used. Open to ideas here, ofc including reverting my change.

Another thought is maybe this is a thing where maybe can log warnings when people try to load this, so that they don't misunderstand why this is here?

chanind:
Yeah might make sense to add a warning so people are aware it's not really meant to used as a standard Transcoder (although the encoder should be fine to use this way). We should be able to write a standard config loader by just getting the header from the safetensors files on Huggingface without fully downloading the whole file, since that breaks CI since it's so big. Mateusz shared some code to do this along these lines:

```
# https://huggingface.co/mntss/clt-131k/resolve/main/W_dec_0.safetensors

import requests
import json
import struct

url = "https://huggingface.co/mntss/clt-131k/resolve/main/W_dec_0.safetensors"
headers = {"Range": "bytes=0-16383"}  # Request first 16KB
response = requests.get(url, headers=headers)
print(f"Status: {response.status_code}")
print(f"Content-Length: {len(response.content)}")

# Parse safetensors metadata
# First 8 bytes contain the length of the JSON metadata
metadata_length = struct.unpack('<Q', response.content[:8])[0]
print(f"Metadata length: {metadata_length}")

# Extract and parse the JSON metadata
metadata_json = response.content[8:8+metadata_length].decode('utf-8')
metadata = json.loads(metadata_json)
print(f"Metadata: {json.dumps(metadata, indent=2)}")

# Extract tensor info
for tensor_name, tensor_info in metadata.items():
    if tensor_name != "__metadata__":  # Skip metadata field if present
        print(f"\nTensor: {tensor_name}")
        print(f"  Shape: {tensor_info['shape']}")
        print(f"  Dtype: {tensor_info['dtype']}")
        print(f"  Data offsets: {tensor_info['data_offsets']}")
 ```
 
 I can edit this PR to add something like this to create a ligher-weight config loader

hijohnnylin:
ah that's very helpful. yes, let's do that!

hijohnnylin:
lgtm! thanks for both adding the warning and getting the lightweight config loader in :D

PR #541, opened 2025-09-02
feat: support Anthropic's JumpReLU training setup

### LINKED ISSUE #427: [Proposal] Support Anthropic's JumpReLU improvements
In https://transformer-circuits.pub/2025/january-update/index.html, Anthropic documents improvements to JumpReLU SAE training. These include the following:

- Using Tanh for the straight-through estimator
- Adding a loss to discourage dead features
- Allowing gradients from the straight-through estimator to flow through to model parameters

We should support these options for JumpReLU training in SAELens

chanind:
# Description

This PR implements Anthropic's JumpReLU setup, based on https://transformer-circuits.pub/2025/january-update/index.html#DL. This has some minor changes from the proposal in #427, specifically:

`jumprelu_ste_method` is instead called `jumprelu_loss_mode`, with options `step` for DeepMind or `tanh` for Anthropic. On further inspection, Anthropic is still technically using a rectangle straight through estimator, but they changed the sparsity loss equation from what Deepmind used. 

I also changed the pre-enc loss to only get applied if SAE latents are actually dead rather than applying any time the latent doesn't fire. It just seems strange to apply a distorting loss to all latents when they're otherwise firing normally IMO.

This PR does not include Anthropic's `b_enc` init method, since when I tested this out it didn't seem to require it to avoide dead features. I suspect that `b_enc` init is likely useful only if the SAE is extremely wide. We can always add that in a follow-up PR. The `b_enc` init method is also likely useable for any SAE, not just JumpReLUs anyway.

I can't find any existing open-source implementation of Anthropic's JumpReLU implementation to compare against, so hopefully this is correct. It seems to work though from what I've experimented with so far. 

Here's a dashboard from a quick test-run on GPT2-small: https://api.wandb.ai/links/chanind/08lb4q3l

Fixes #427 

## Type of change

### DIFF
--- a/docs/training_saes.md
+++ b/docs/training_saes.md
@@ -154,6 +154,37 @@ sparse_autoencoder = LanguageModelSAETrainingRunner(cfg).run()
 
 [JumpReLU SAEs](https://arxiv.org/abs/2407.14435) are a state-of-the-art SAE architecture. To train one, provide a `JumpReLUTrainingSAEConfig` to the `sae` field. JumpReLU SAEs use a sparsity penalty controlled by the `l0_coefficient` parameter. The `JumpReLUTrainingSAEConfig` also has parameters `jumprelu_bandwidth` and `jumprelu_init_threshold` which affect the learning of the thresholds.
 
+We support both the original JumpReLU sparsity loss and the more modern [tanh sparsity loss](https://transformer-circuits.pub/2025/january-update/index.html) variant from Anthropic. To use the tanh sparsity loss, set `jumprelu_sparsity_loss_mode="tanh"`. The tanh sparsity loss variant is a bit easier to train, but has more hyper-parameters. We recommend using the tanh with `normalize_activations="expected_average_only_in"` to match what Anthropic's setup. We also recommend enabling the pre-act loss by setting `pre_act_loss_coefficient` to match Anthropic's setup. An example of this is below:
+
+```python
+from sae_lens import LanguageModelSAERunnerConfig, LanguageModelSAETrainingRunner, JumpReLUTrainingSAEConfig
+
+cfg = LanguageModelSAERunnerConfig( # Full config would be defined here
+    # ... other LanguageModelSAERunnerConfig parameters ...
+    sae=JumpReLUTrainingSAEConfig(
+        l0_coefficient=5.0, # Sparsity penalty coefficient
+        jumprelu_sparsity_loss_mode="tanh",
+        jumprelu_tanh_scale=4.0, # default value
+        jumprelu_bandwidth=2.0,
+        jumprelu_init_threshold=0.1,
+        pre_act_loss_coefficient=3e-6,
+        # Anthropic's settings assume normalized activations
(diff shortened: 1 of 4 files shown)


docs/training_saes.md
```diff
@@ -154,6 +154,37 @@ sparse_autoencoder = LanguageModelSAETrainingRunner(cfg).run()
 
 [JumpReLU SAEs](https://arxiv.org/abs/2407.14435) are a state-of-the-art SAE architecture. To train one, provide a `JumpReLUTrainingSAEConfig` to the `sae` field. JumpReLU SAEs use a sparsity penalty controlled by the `l0_coefficient` parameter. The `JumpReLUTrainingSAEConfig` also has parameters `jumprelu_bandwidth` and `jumprelu_init_threshold` which affect the learning of the thresholds.
 
+We support both the original JumpReLU sparsity loss and the more modern [tanh sparsity loss](https://transformer-circuits.pub/2025/january-update/index.html) variant from Anthropic. To use the tanh sparsity loss, set `jumprelu_sparsity_loss_mode="tanh"`. The tanh sparsity loss variant is a bit easier to train, but has more hyper-parameters. We recommend using the tanh with `normalize_activations="expected_average_only_in"` to match what Anthropic's setup. We also recommend enabling the pre-act loss by setting `pre_act_loss_coefficient` to match Anthropic's setup. An example of this is below:
```
anthonyduong9, line 157:
```suggestion
We support both the original JumpReLU sparsity loss and the more modern [tanh sparsity loss](https://transformer-circuits.pub/2025/january-update/index.html) variant from Anthropic. To use the tanh sparsity loss, set `jumprelu_sparsity_loss_mode="tanh"`. The tanh sparsity loss variant is a bit easier to train, but has more hyper-parameters. We recommend using the tanh with `normalize_activations="expected_average_only_in"` to match Anthropic's setup. We also recommend enabling the pre-act loss by setting `pre_act_loss_coefficient` to match Anthropic's setup. An example of this is below:
```

sae_lens/saes/jumprelu_sae.py
```diff
@@ -310,3 +352,21 @@ def process_state_dict_for_loading(self, state_dict: dict[str, Any]) -> None:
...
+    pre_act_loss_coefficient: float,
+    threshold: torch.Tensor,
+    hidden_pre: torch.Tensor,
+    dead_neuron_mask: torch.Tensor | None,
+    W_dec_norm: torch.Tensor,
+) -> torch.Tensor:
+    """
+    Calculate Anthropic's pre-activation loss, except we only calculate this for latents that are actually dead.
+    """
+    if dead_neuron_mask is None or dead_neuron_mask.sum() == 0:
```
anthonyduong9, line 367: If we want to trust (or guarantee) that `dead_neuron_mask` only contains `0`s and `1`s, we can replace `dead_neuron_mask.sum() == 0` with `not dead_neuron_mask.any()`, which Claude Code and ChatGPT 5 Thinking say is usually faster.

PR #547, opened 2025-09-08
fix: allow float for batchtopk k param

chanind:
# Description

Since BatchTopK ensures that a mean k across a batch rather than ensuring that every sample in the batch has a set k, it can operate with a floating point number for k without issue. E.g., it's fine to set `k = 1.5` for BatchTopK, and for a batch of 10 items, there will be 15 non-zero activations across the batch. However, our current implementation types the k param as an `int` instead of a `float.

This PR updates the types for BatchTopK to accept floats, and adds tests to ensure this works as expected.

## Type of change

### DIFF
--- a/sae_lens/saes/batchtopk_sae.py
+++ b/sae_lens/saes/batchtopk_sae.py
@@ -15,7 +15,7 @@ class BatchTopK(nn.Module):
 
     def __init__(
         self,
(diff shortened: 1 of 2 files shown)


chanind:
merging as this should be uncontroversial

PR #550, opened 2025-09-19
opt: keep post topK activations as a sparse COO matrix & use smm in decode

wz-ml:
# Description

This PR adds support for [sparse COO tensors](https://docs.pytorch.org/docs/2.8/sparse.html) as the intermediate activations for the topK SAE.

Additions:
- [x] Support for sparse COO tensors.
- [x] Transparency for hooks (e.g if hooks are applied, the intermediate tensors will (1) appear to be the right shape and (2) will be densified before operations are applied to them).
- [x] Added tests.
- [x] Added benchmark script for performance replication.

For speedups on my machine (3090), see below:

image

## Type of change

# Checklist:


sae_lens/saes/topk_sae.py
```diff
@@ -53,6 +103,7 @@ class TopKSAEConfig(SAEConfig):
     """
 
     k: int = 100
+    sparse_intermediate: bool = True
```
chanind, line 106: nit: could this be named something like `use_sparse_tensors`? the name `sparse_intermediate` doesn't feel obvious what it means

sae_lens/saes/topk_sae.py
```diff
@@ -93,24 +147,65 @@ def encode(
...
+            return self.hook_sae_acts_post(
+                self.activation_fn(hidden_pre), x_shape=x.shape
+            )
         return self.hook_sae_acts_post(self.activation_fn(hidden_pre))
 
     def decode(
-        self, feature_acts: Float[torch.Tensor, "... d_sae"]
+        self,
+        feature_acts: Float[torch.Tensor, "... d_sae"],
+        x_shape: torch.Size | None = None,
```
chanind, line 162: Why is this shape param needed? We can't just call `.shape` on the actual tensor itself? I'd rather avoid changing the function definition of `decode()` unless it's absolutely necessary. If the sparse tensors version of TopK only works with certain shaped tensors that's OK IMO, as long as we output an understandable error message to the user and make it easy for them to disable the sparse implementation (e.g. calling `sae.disable_sparse_tensors()` or `sae.use_sparse_tensors(False)` or something.

tests/saes/test_topk_sae.py
```diff
@@ -146,3 +146,77 @@ def test_TopKTrainingSAE_save_and_load_inference_sae(tmp_path: Path) -> None:
...
+    # as when using the dense intermediates.
+    d_in = 128
+    d_sae = 1024
+    M = 128
+
+    cfg = build_topk_sae_training_cfg(
+        d_in=d_in,
+        d_sae=d_sae,
+        k=26,
+        decoder_init_norm=1.0,  # TODO: why is this needed??
```
chanind, line 180: Why is this needed? It errors if you remove it?

sae_lens/saes/sae.py
```diff
@@ -14,7 +14,6 @@
     Generic,
     Literal,
     NamedTuple,
-    Type,
```
chanind, line 17: 👍

sae_lens/saes/topk_sae.py
wz-ml: Thanks, good point - I'll rename this.

sae_lens/saes/topk_sae.py
wz-ml: Yeah, this is not ideal and I'm happy to consider other ways to refactor this. The reason why I pass in this parameter's because:
1. In the decode function, we have this matmul: `sae_out_pre = feature_acts @ self.W_dec + self.b_dec`. In the case where feature_acts has shape [..., d_in] (more than 1 batch dim), torch calls a reshape on feature_acts to flatten the batch dims and reshape the output.
2. Sparse COO tensors don't support reshaping after creation, and torch errors out when it tries to reshape to do the above.

Yeah, sounds good - if 2D input shapes are the expected default, we can have this error out and make this implementation easily toggle-able via a function call.

tests/saes/test_topk_sae.py
wz-ml: Good catch, removed.

sae_lens/saes/sae.py
wz-ml: Thanks!

chanind:
Summary of offline discussion about this PR:

- It turns out that `torch.sparse.mm` does not work with bfloat16 tensors, and thus doesn't support autocasting. Having to disable autocasting for this operation results in a larger performance hit than the efficiency gain of using sparse tensors. As a result, this optimization only makes sense if the user can't use bfloat16 autocasting, likely due to being on old hardware.
- We will still merge this, but default it to disabled until we can find a better way to handle sparse decoding in topk SAEs.

PR #561, opened 2025-10-06
Resume training from checkpoints

chanind:
# Description

This PR allows resuming training from checkpoints. Currently, we save checkpoints during training, but we cannot easily resume training from these checkpoints, limiting their usefulness. This PR does the following:

- Saves full trainer state with each checkpoint so training can resume exactly as it left off
- Enables loading activation store from saved checkpoint, so that the store will fast-forward to the target location in the dataset.
- Allows passing a `resume_checkpoint_path` param to the LLM Training Runner both on CLI and in code to resume training from a saved checkpoint.

## Type of change

# Checklist:


sae_lens/training/sae_trainer.py
```diff
@@ -93,11 +101,6 @@ def __init__(
             sae.cfg.d_sae, device=cfg.device
         )
         self.n_frac_active_samples = 0
-        # we don't train the scaling factor (initially)
-        # set requires grad to false for the scaling factor
-        for name, param in self.sae.named_parameters():
-            if "scaling_factor" in name:
```
chanind, line 99: scaling is no longer handled in the SAE, cleaning this up

sae_lens/training/sae_trainer.py
```diff
@@ -84,7 +90,9 @@ def __init__(
                 range(
                     0,
                     cfg.total_training_samples,
-                    cfg.total_training_samples // self.cfg.n_checkpoints,
+                    math.ceil(
+                        cfg.total_training_samples / (self.cfg.n_checkpoints + 1)
```
chanind, line 94: There was some wierdness happening when `total_training_samples` was not exactly divisible by `n_checkpoints`, where you could end up with an extra checkpoint. This should fix that issue.

sae_lens/llm_sae_training_runner.py
```diff
@@ -315,7 +322,7 @@ def save_checkpoint(
 
 def _parse_cfg_args(
     args: Sequence[str],
-) -> LanguageModelSAERunnerConfig[TrainingSAEConfig]:
+) -> tuple[LanguageModelSAERunnerConfig[TrainingSAEConfig], str | None]:
```
anthonyduong9: What do you think of adding a `resume_from_checkpoint` field to `LanguageModelSAERunnerConfig` and reverting the changes to `_parse_cfg_args()`?

PR #605, opened 2025-12-24
feat: Add Swanlab logging backend

### LINKED ISSUE #604: [Proposal] Support SwanLab as an alternative logging backend to wandb
### Proposal

Add optional SwanLab logging support to SAELens.

### Motivation

Some users cannot reliably use Weights & Biases (wandb) due to network restrictions. SwanLab can serve as a drop-in alternative for experiment logging in these environments.

### Pitch

1. Add `wandb_compat.py`, which selects `swanlab` or `wandb` based on `SAE_LENS_LOGGING_BACKEND`, and provides a fallback `generate_id()` implementation (use UUID when `wandb.util.generate_id` is unavailable).
2. Keep calling `wandb.init()` / `wandb.finish()` in `llm_sae_training_runner.py`, but route the actual backend module through the compatibility layer (either SwanLab or wandb).
3. When the backend is SwanLab, skip Artifact-related `log_artifact` calls in `config.py` (to avoid unsupported features), and stop using `wandb.Histogram` in `sae_trainer.py`.

### Checklist

* [x] I have checked that there is no similar [[issue](https://github.com/decoderesearch/SAELens/issues)](https://github.com/decoderesearch/SAELens/issues) in the repo (**required**)

wantbook-book:
# Description

Summary:

- Add a logging compatibility layer to support SwanLab or W&B backends and disable Artifact/Histogram logging when using SwanLab.
- Add a minimal SwanLab logging test entry point and pytest guard to validate a tiny StandardSAE run.

Motivation:

- SwanLab does not support W&B Artifacts/Histogram objects, which caused logging failures. The change routes logging through a backend selector and conditionally disables unsupported features while preserving scalar logs and training flow.

Dependencies:

- `swanlab` (when `SAE_LENS_LOGGING_BACKEND=swanlab`)

Fixes #604 

## Type of change

# Checklist:

- [ ] New and existing tests pass locally with my changes

### Performance Check.

- [ ] L0
- [ ] CE Loss
- [ ] MSE Loss

### DIFF
--- a/tests/_comparison/sae_lens_old/config.py
+++ b/tests/_comparison/sae_lens_old/config.py
@@ -6,6 +6,7 @@ from typing import Any, Literal, cast
 
 import simple_parsing
 import torch
+import wandb
 from datasets import (
     Dataset,
     DatasetDict,
@@ -14,7 +15,6 @@ from datasets import (
     load_dataset,
 )
 
-import wandb
 from tests._comparison.sae_lens_old import __version__, logger
 
 DTYPE_MAP = {
--- a/tests/_comparison/sae_lens_old/sae_training_runner.py
+++ b/tests/_comparison/sae_lens_old/sae_training_runner.py
@@ -6,10 +6,10 @@ from pathlib import Path
 from typing import Any, cast
 
 import torch
+import wandb
 from simple_parsing import ArgumentParser
 from transformer_lens.hook_points import HookedRootModule
 
-import wandb
 from tests._comparison.sae_lens_old import logger
 from tests._comparison.sae_lens_old.config import (
     HfDataset,
--- a/tests/_comparison/sae_lens_old/training/sae_trainer.py
+++ b/tests/_comparison/sae_lens_old/training/sae_trainer.py
@@ -2,11 +2,11 @@ import contextlib
 from typing import Any, Protocol, cast
 
 import torch
+import wandb
 from torch.optim import Adam
 from tqdm import tqdm
 from transformer_lens.hook_points import HookedRootModule
 
-import wandb
 from tests._comparison.sae_lens_old import __version__
 from tests._comparison.sae_lens_old.config import LanguageModelSAERunnerConfig
 from tests._comparison.sae_lens_old.evals import EvalConfig, run_evals


wantbook-book:
1. CI is failing because the workflow attempts to request an OIDC token, but `permissions: id-token: write` isn’t enabled for the job/workflow. Since this PR is from a fork, token/permission restrictions may also apply.

2. For linting, I ran `poetry run ruff check --fix tests/_comparison/sae_lens_old` locally and got `All checks passed!`.

sae_lens/wandb_compat.py
```diff
@@ -0,0 +1,39 @@
...
+import uuid
+from typing import Any
+
+
+def _load_backend() -> tuple[Any, str]:
+    backend = os.getenv("SAE_LENS_LOGGING_BACKEND", "auto").lower()
+    if backend not in {"auto", "wandb", "swanlab"}:
+        backend = "auto"
+
+    if backend in {"auto", "swanlab"}:
```
chanind: I'd rather the default be wandb, not swanlab. If a user just happens to have SAELens and swanlab deps in the same project their logging will seem to break and this will be very hard to debug. IMO if users want to use swanlab they should need to opt-in specifically with a ENV var.

sae_lens/training/sae_trainer.py
```diff
@@ -398,8 +398,9 @@ def _run_and_log_evals(self):
                 if self.evaluator is not None
                 else {}
             )
-            for key, value in self.sae.log_histograms().items():
-                eval_metrics[key] = wandb.Histogram(value)  # type: ignore
+            if BACKEND != "swanlab":
```
chanind: I'd rather not need to have if-statements like this in code. Can you provide an interface so swanlab works identical to wandb, and we don't need any of these checks? Or maybe make a generic `Logger` interface and have both Swanlab and WandB versions of that same interface? It shouldn't be the responsibility of the SAETrainer to keep track of what logger is being used.

tests/test_swanlab_log.py
```diff
@@ -0,0 +1,65 @@
...
+
+def main() -> None:
+    os.environ.setdefault("SAE_LENS_LOGGING_BACKEND", "swanlab")
+    os.environ.setdefault("SWANLAB_MODE", "offline")
+    os.environ.setdefault("WANDB_MODE", "offline")
+    output_dir = Path(os.getenv("SAE_OUTPUT_DIR", "swanlab_minimal_output"))
+    run_minimal_standard_sae_training(output_dir)
+
+
+if __name__ == "__main__":
```
chanind: nit: don't use main functions in tests. Just run the test using pytest.

chanind (commented):
Thank you for contributing this! However, I'm inclined to hold off on merging this until the next major release of SAELens. Ideally we should have a generic "Logger" interface, and then the user should specify the loggers they want in the config. That should make it possible for users to register their own loggers if they want as well. I would envision something like the following:

```python
  ...
  logger=[
      WandBLoggingConfig(...),
      SwanLabLoggingConfig(...),
  ],
)
```

This would be a breaking change though so we'd need to de a major release, and requires a bit more refactoring than is ideal in just this PR.

PR #608, opened 2025-12-28
fix: loading of some batchtopk SAEs from pretrained.

shawnghu:
Prior to this commit, the following code:

```
import torch
from sae_lens import SAE
from sae_lens.analysis.hooked_sae_transformer import
HookedSAETransformer

release, sae_id, model_name = "llama-3.1-8b-instruct-andyrdt",
"resid_post_layer_23_trainer_1", "meta-llama/Llama-3.1-8B-Instruct"
sae = SAE.from_pretrained(release, sae_id, device='cpu')
```

would fail with the error:

```
    RuntimeError: Error(s) in loading state_dict for StandardSAE:
        Unexpected key(s) in state_dict: "threshold".
```
It seems likely that many other BatchTopKSAEs, particularly those by the same author, would suffer from the same.

Below is a partial history of this issue generated by Claude Code and then perused by me for correctness and edited by me for brevity.

The andyrdt SAEs use `dict_class: BatchTopKSAE` in their config, which in particular has a field named "threshold". However, the architecture detection in `get_dictionary_learning_config_1_from_hf` only recognized `MatryoshkaBatchTopKSAE` as needing the `jumprelu` architecture, not plain `BatchTopKSAE`.

1. **Before commit 58d0cd7** (SAEBench Matryoshka SAEs):
   - The loader did not add `threshold` to state_dict
   - BatchTopKSAE loaded as StandardSAE with architecture="standard"
   - The threshold parameter was silently discarded
   - SAEs loaded successfully but **incorrectly**: activations were computed with plain ReLU instead of threshold-masked JumpReLU
   - This meant wrong latent activations, but no error

2. **After commit 58d0cd7**:
   - The commit added threshold handling: `if "threshold" in encoder: state_dict["threshold"] = threshold`
   - It also added architecture detection for `MatryoshkaBatchTopKSAE` -> `jumprelu`
   - But `BatchTopKSAE` still fell through to `architecture: standard`
   - Now the loader includes `threshold` in state_dict, but StandardSAE rejects the unexpected key
   - SAEs fail to load entirely

There's a remark in the batchtopk training that they are saved as jumprelu SAEs for inference, which seems to have been severely misleading for these. My guess about this (which the repo owners should know the ground truth about) is that the andyrdt SAEs were trained not using this repo, so their conversion to JumpReLU and inclusion of 'threshold' parameter was simply never supported in the first place.

# Description

Please include a summary of the change and which issue is fixed. Please also include relevant motivation and context. List any dependencies that are required for this change.

Fixes # (issue)

## Type of change

# Checklist:

### Performance Check.

- [ ] L0
- [ ] CE Loss
- [ ] MSE Loss

### DIFF
--- a/tests/loading/test_pretrained_sae_loaders.py
+++ b/tests/loading/test_pretrained_sae_loaders.py
@@ -17,6 +17,7 @@ from sae_lens.loading.pretrained_sae_loaders import (
     gemma_2_transcoder_huggingface_loader,
     gemma_3_sae_huggingface_loader,
     get_deepseek_r1_config_from_hf,
+    get_dictionary_learning_config_1_from_hf,
     get_gemma_2_transcoder_config_from_hf,
     get_gemma_3_config_from_hf,
     get_goodfire_config_from_hf,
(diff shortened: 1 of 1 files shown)


shawnghu:
I'd like some input from the maintainers about what to do in terms of unit tests for this. Or, if you want to merge this immediately as a hotfix and then do the unit test stuff yourself, that's fine.

I believe documentation will now be *correct* as a result of this change, so technically no documentation changes are needed.

chanind:
Ack great catch with this! I added a test that the config at least is loaded as jumprelu for the andyrdt SAE. Longer-term, maybe we can add a test that uses the actual `dictionary_learning` library and saves out an SAE of each type and verifies it can be loaded properly in SAELens. For now though, I think the config test should be good enough. I'll merge this once tests pass.

chanind:
Also great job with figuring out the commit history of where this bug came from. It definitely sounds like this was silently causing problems for awhile now.

PR #640, opened 2026-02-06
feat: add SwanLab as alternative logging backend

### LINKED ISSUE #604: [Proposal] Support SwanLab as an alternative logging backend to wandb

puranikyashaswin:
# Description

Adds SwanLab as an optional alternative logging backend to wandb for users with network restrictions. This implements a compatibility layer that abstracts the logging backend, allowing users to switch between wandb and SwanLab via an environment variable.

Fixes #604

## Type of change

# Checklist:

### Performance Check.

Not applicable - this is a logging infrastructure change only, no training behavior is modified.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # CHANGELOG
 
+## v6.41.2 (2026-04-27)
+
+### Fix
+
+* fix: more cross-device tensor bugs in evals (#676) ([`9311085`](https://github.com/decoderesearch/SAELens/commit/93110856b0c0bd6a6dcb0a2d45a01cba5a2272ca))
+
 ## v6.41.1 (2026-04-27)
 
 ### Fix
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "sae-lens"
-version = "6.41.1"
+version = "6.41.2"
 description = "Training and Analyzing Sparse Autoencoders (SAEs)"
 authors = ["Joseph Bloom"]
 readme = "README.md"
--- a/sae_lens/__init__.py
+++ b/sae_lens/__init__.py
@@ -1,5 +1,5 @@
 # ruff: noqa: E402
-__version__ = "6.41.1"
+__version__ = "6.41.2"
 
 import logging
 


chanind:
This is the second PR adding swanlab after #605, so I'll post the same response as to that PR here:

Thank you for contributing this! However, I'm inclined to hold off on merging this until the next major release of SAELens. Ideally we should have a generic "Logger" interface, and then the user should specify the loggers they want in the config. That should make it possible for users to register their own loggers if they want as well. I would envision something like the following:

```python
  ...
  logger=[
      WandBLoggingConfig(...),
      SwanLabLoggingConfig(...),
  ],
)
```

This would be a breaking change though so we'd need to de a major release, and requires a bit more refactoring than is ideal in just this PR. I'll aim to get something like this in the next few weeks.

puranikyashaswin:
@chanind Thanks for the thoughtful feedback! The generic Logger interface is a much cleaner approach I can see how it would make the codebase more maintainable and give users flexibility.

I'd be glad to help out with the refactoring whenever you start working on it. Happy to take on whatever part would be most useful whether that's the interface design, migrating existing WandB logic, or adding SwanLab support after.

Should I close this PR for now, or leave it open as a reference? Either way, feel free to tag me when the work kicks off.

chanind:
let's just leave it open for now, I'll let you know once the refactor of the logging is further along! Thank you again for this!

PR #665, opened 2026-03-26
Fix explained_variance computing variance relative to zero instead of mean

amaljithkuttamath:
## Summary

Fixes #659. Two bugs in `get_sparsity_and_variance_metrics` caused `explained_variance` to compute total variance relative to zero instead of relative to the mean:

1. **Line 552**: `mean_act_per_dimension` accumulated `.pow(2).mean(dim=0)` (i.e. `E[x_d^2]`) instead of `.mean(dim=0)` (i.e. `E[x_d]`). When squared on line 586, this produced `E[x_d^2]^2` instead of the correct `E[x_d]^2`.

2. **Line 585**: `torch.cat` on a list of `(d_model,)` tensors produced `(N_batches * d_model,)`, then `.mean(dim=0)` collapsed everything to a scalar, destroying per-dimension structure. Replaced with `torch.stack` and added `.sum()` to reduce across dimensions after squaring (matching how `mean_sum_of_squares` is already a scalar from `.sum(dim=-1)`).

The combined effect: `total_variance` was approximately `E[||X||^2]` (variance from zero) instead of `E[||X||^2] - ||E[X]||^2` (variance from mean). For activations with large mean components, this inflated `explained_variance`.

## Changes

- `sae_lens/evals.py`: 3 line changes (remove `.pow(2)`, `cat` -> `stack`, add `.sum()`)
- `tests/test_evals.py`: New test that constructs high-mean data and verifies the variance formula matches `torch.var` with `correction=0`. Also verifies the buggy formula produces a materially different (incorrect) result.

## Test plan

- [x] New test `test_explained_variance_uses_mean_centered_variance` passes
- [x] Existing test `test_get_sparsity_and_variance_metrics_identity_sae_perfect_reconstruction` still passes
- [x] `ruff check` passes


chanind:
Thank you for this PR. This PR is probably correct, but I just want to make sure before merging to avoid needing to fix this calculation a third time.

tests/test_evals.py
```diff
@@ -644,6 +644,53 @@ def test_get_sparsity_and_variance_metrics_identity_sae_perfect_reconstruction(
...
+    # With the bug (.pow(2) on the mean term), the subtracted term becomes
+    # sum(E[x_d^2]^2) instead of sum(E[x_d]^2). For large-mean data this
+    # makes buggy_total_var very negative (or wildly wrong in general),
+    # which distorts the explained_variance ratio.
+    buggy_mean_act = x.pow(2).mean(dim=0)  # bug: .pow(2) before mean
+    buggy_total_var = (mean_sum_of_squares - (buggy_mean_act**2).sum()).item()
+    assert abs(buggy_total_var - expected_total_var) > expected_total_var * 0.5
+
+
+def test_explained_variance_zero_total_variance():
```
chanind, line 751: This test doesn't actually test any code.

tests/test_evals.py
```diff
@@ -644,6 +644,53 @@ def test_get_sparsity_and_variance_metrics_identity_sae_perfect_reconstruction(
     assert metrics["mse"] == pytest.approx(0.0, abs=1e-5)
 
 
+def test_explained_variance_uses_mean_centered_variance():
```
chanind: This test doesn't test any actual code

chanind (commented):
Main issues are the tests do not actually test anything, I'll try working on this

robbiebusinessacc:
Came across this digging through the eval code, and since the thread mentions
  wanting extra confidence before merging a third time, here's an independent
  hand-computed check of #665's fix that doesn't lean on `torch.var` as the
  oracle:

  Two tokens, two dims:

  x      = [[2, 0], [0, 4]]   ->  mu = [1, 2]
  x_hat  = [[2, 1], [0, 3]]
  total_variance     = E[Σ_d (x_d - mu_d)²] = ((1+4) + (1+4)) / 2 = 5
  residual_variance  = E[Σ_d (x_d - x̂_d)²] = (1 + 1) / 2 = 1
  explained_variance = 1 - 1/5 = 0.8

  I ran #665's code on this and it returns `0.8`; the pre-fix code does not. Two
  cheap canaries that would also lock it down going forward, if useful as extra
  test cases:

  - **Predict-the-mean** (`x_hat = x.mean(0)`): EV must be ~0. The old code
  returns a value > 1, which is impossible — a clean regression canary.
  - **Perfect reconstruction** (`x_hat = x`): EV must be exactly 1.

chanind:
I've verified this fix carefully against the original issue (#659) and pushed an update (03c637ca) that rebases onto current main and restructures the implementation.

**Verification of the math.** The diagnosis in #659 is correct: the old code squared `E[x_d²]` instead of `E[x_d]`, and `torch.cat(...).mean()` collapsed the per-dimension means to a scalar, so `total_variance ≈ E[||x||²]` — variance relative to a zero baseline. The fix in this PR computed the intended `Var(X) = E[||X||²] - ||E[X]||²` correctly for equal-sized batches.

**Two remaining subtleties fixed in the update:**

1. **Unequal batch weighting.** Eval batches are masked with `ignore_tokens` before being accumulated, so they contain different numbers of tokens. Averaging per-batch means weights small batches the same as large ones, which biases all three statistics. The update accumulates running sums over samples instead, so every token is weighted equally regardless of how batches are sized.

2. **Duplicate implementation.** `sae_lens.synthetic.evals.ExplainedVarianceCalculator` already implemented this metric correctly as a streaming calculator. Rather than maintaining two implementations of the same formula, the calculator now lives in `sae_lens.evals` and is used by both the LLM eval loop and the synthetic evals (the synthetic module re-imports it). This also removes the memory cost of retaining all activations across eval batches, and the cross-batch sums now accumulate in float64 to reduce catastrophic cancellation in `E[||x||²] - ||E[x]||²` for high-mean activations.

**Tests** now include the canaries @robbiebusinessacc suggested (thanks — the hand-computed example is in there verbatim): predict-the-mean must give EV ≈ 0, perfect reconstruction must give EV = 1, plus a `torch.var` oracle check, exact batched-vs-unbatched equality with *unequal* batch sizes, bias invariance in float64, zero-variance edge cases, and a known-noise-level statistical check (EV = 0.75 for noise_std = 0.5 on unit-variance data). There's also an end-to-end test that shifting activations by a constant (with `b_dec` adjusted to match) leaves `explained_variance` unchanged.

Note: #710 is a duplicate fix for the same issue and can be closed if this merges.

PR #668, opened 2026-04-16
Update link for pre-trained SAEs in READMEs

hijohnnylin:
# Description
Seems that the link currently leads to an image? Added correct link.

## Type of change


chanind (approved):
LGTM!

PR #677, opened 2026-04-30
feat: add Qwen Scope TopK SAEs

chanind:
## Summary

Adds support for the 14 [Qwen Scope](https://huggingface.co/collections/Qwen/qwen-scope) TopK SAE releases — 552 SAEs across 7 Qwen3 / Qwen3.5 backbones (dense and MoE), residual stream, all transformer layers.

- New `qwen_scope` loader (`get_qwen_scope_config_from_hf` + `qwen_scope_sae_huggingface_loader`) that parses the SAE repo name (`SAE-Res-{model}-W{N}K-L0_{K}`) and per-layer filename (`layer{n}.sae.pt`) to derive `d_in`, `d_sae`, `k`, `model_name`, `hook_name`, and `hf_hook_name`. Transposes `W_enc` (d_sae,d_in) and `W_dec` (d_in,d_sae) into SAELens convention.
- 14 YAML releases (Qwen3 1.7B/8B/30B-A3B-Base, Qwen3.5 2B/9B/35B-A3B-Base, Qwen3.5-27B), each at L0 ∈ {50, 100} and width 16× / 64× hidden_size. Per the [technical report](https://qianwen-res.oss-accelerate.aliyuncs.com/qwen-scope/Qwen_Scope.pdf) Table 1, all backbones are the `-Base` variant except Qwen3.5-27B, which is trained on the instruct variant. SAE ids use the `layer{n}` form to mirror the source path.
- `dataset_path` is left as `None` since Qwen Scope is trained on Qwen's in-house pretraining data, which is not public.

## Test plan

- [x] `poetry run pytest tests/loading/test_pretrained_sae_loaders.py` (67 passed, 1 unrelated skip)
- [x] `poetry run ruff check` clean on changed files
- [x] `poetry run pyright` clean on the loader file
- [x] 8 new tests covering: config parsing for both Qwen3 and Qwen3.5 variants, overrides, three error cases, end-to-end weight transpose correctness (verifies `residual @ W_enc.T + b_enc` matches SAELens `residual @ W_enc + b_enc` after the transpose), and a full `SAE.from_dict` + `sae.encode(...)` reproducing the README's reference TopK formula.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/sae_lens/loading/pretrained_sae_loaders.py
+++ b/sae_lens/loading/pretrained_sae_loaders.py
@@ -1134,7 +1134,6 @@ def get_qwen_scope_config_from_hf(
         # Qwen Scope SAEs are trained on Qwen's in-house pretraining data, which
         # is not publicly available; leave the dataset_path unset.
         "dataset_path": None,
-        "dataset_trust_remote_code": True,
         "sae_lens_training_version": None,
         "normalize_activations": "none",
         "device": device,
--- a/sae_lens/loading/pretrained_sae_loaders.py
+++ b/sae_lens/loading/pretrained_sae_loaders.py
@@ -1067,6 +1067,121 @@ def deepseek_r1_sae_huggingface_loader(
     return cfg_dict, state_dict, None
 
 
+# Maps the model identifier embedded in a Qwen Scope SAE repo name (the part
+# between "SAE-Res-" and "-W{N}K-L0_{K}") to the HF model id and its hidden size.
+QWEN_SCOPE_MODEL_INFO: dict[str, tuple[str, int]] = {
+    "Qwen3-1.7B-Base": ("Qwen/Qwen3-1.7B-Base", 2048),
+    "Qwen3-8B-Base": ("Qwen/Qwen3-8B-Base", 4096),
+    "Qwen3-30B-A3B-Base": ("Qwen/Qwen3-30B-A3B-Base", 2048),
+    "Qwen3.5-2B-Base": ("Qwen/Qwen3.5-2B-Base", 2048),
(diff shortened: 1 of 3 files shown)


chanind:
Thanks for the review! Quick rundown:

- **Qwen3.5-27B model name** — verified that `Qwen/Qwen3.5-27B-Instruct` does not exist on HF (returns 401, only `Qwen/Qwen3.5-27B` is published). For the Qwen3.5 series, the bare-name variant *is* the post-trained/instruct model and the `-Base` suffix marks the pretraining-only variant (same pattern holds for Qwen3.5-2B / -9B / -35B-A3B). So `Qwen/Qwen3.5-27B` is the correct id for the instruct backbone — leaving as-is.
- **Test docstring** — removed in 280d3fb.
- **`device="cuda"` in CPU-only test** — switched to `"cpu"` in 280d3fb (the assertion was tautological anyway since the config getter doesn't move tensors).

`weights_only=True` for `torch.load` is a real follow-up but applies to the whole loaders module, so out of scope for this PR.

chanind:
Round 2 follow-ups:

- **`dataset_trust_remote_code` with `dataset_path: None`** — good catch, removed in 51ba5c4.
- **`context_size: 1024`** — keeping as-is; the technical report's "Training in Practice" (Section 2.2) doesn't specify a training context length.
- **MoE W32K vs W128K release pairing** — accurate observation but not a defect, just upstream's design (per Table 1, "for MoE models, we additionally release wider SAEs to capture more fine-grained features"). Worth noting for users in the docs but no code change.
- **`torch.load(weights_only=True)`** — acknowledged, codebase-wide cleanup, out of scope.

PR #691, opened 2026-06-06
Fix incorrect buffer size calculations in ActivationsStore 

danra:
# Description

`ActivationsStore.get_data_loader` and `.get_multi_hook_data_loader` both calculated the buffer size incorrectly, which I stumbled upon via an assertion error running `test_language_model_sae_runner_othellogpt` under `benchmark/test_language_model_sae_runner.py` - which doesn't run in CI which is probably why it went undetected.

## Type of change

- [x] (Somewhat of a) Breaking change (fix or feature that would cause existing functionality to not work as expected)

# Checklist:

- [x] I have ~added~ modified a test to prove my fix is effective

### Performance Check.

Not done; if deemed necessary, I'd appreciate some guidance, i.e., what to use for control and test groups.

- [ ] L0
- [ ] CE Loss
- [ ] MSE Loss

### DIFF
--- a/sae_lens/training/activations_store.py
+++ b/sae_lens/training/activations_store.py
@@ -821,7 +821,7 @@ class ActivationsStore:
         Return an auto-refilling stream of filtered and mixed activations.
         """
         return mixing_buffer(
-            buffer_size=self.n_batches_in_buffer * self.training_context_size,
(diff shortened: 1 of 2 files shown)


danra:
Pulling back the PR, don't think the fix is right

PR #692, opened 2026-06-06
Fix incorrect buffer size calculations in ActivationsStore

danra:
## Description

`ActivationsStore.get_data_loader` and `.get_multi_hook_data_loader` both calculated the buffer size incorrectly, which I stumbled upon via an assertion error running `test_language_model_sae_runner_othellogpt` under `benchmark/test_language_model_sae_runner.py` - which doesn't run in CI which is probably why it went undetected.

A couple of other tangential issues also came up in the way to fix the main problem, and are also fixed here.

## Review Guidance

Best reviewed commit-by-commit.

## Type of change

- [x] Breaking change (fix or feature that would cause existing functionality to not work as ~expected~ before) - buffer size would now be larger in existing code.

# Checklist:

- [ ] I have not made corresponding changes to the documentation - is this significant enough to be documented as a breaking change? Performance might differ (see below) as well as mixing behavior (which would now work as it was supposed to before)

### Performance Check.

Tested `benchmark/test_language_model_sae_runner.py:test_language_model_sae_runner_top_k`, losses seems unchanged, but on my machine the test now takes more time (~20secs vs previous ~15secs) - possibly due to the larger memory footprint.

### DIFF
--- a/sae_lens/training/activations_store.py
+++ b/sae_lens/training/activations_store.py
@@ -821,7 +821,9 @@ class ActivationsStore:
         Return an auto-refilling stream of filtered and mixed activations.
         """
         return mixing_buffer(
-            buffer_size=self.n_batches_in_buffer * self.training_context_size,
+            buffer_size=self.n_batches_in_buffer
+            * self.store_batch_size_prompts
+            * self.training_context_size,
             batch_size=self.train_batch_size_tokens,
             activations_loader=self._iterate_filtered_activations(),
             mix_fraction=self.activations_mixing_fraction,
@@ -943,7 +945,9 @@ class ActivationsStore:
                 "via from_config_multi_hook"
             )
         return multi_hook_concat_split_iter(
-            buffer_size=self.n_batches_in_buffer * self.training_context_size,
+            buffer_size=self.n_batches_in_buffer
+            * self.store_batch_size_prompts
+            * self.training_context_size,
             batch_size=self.train_batch_size_tokens,
             activations_loader=self._iterate_filtered_multi_hook_activations(),
             hook_names=list(self._hook_names),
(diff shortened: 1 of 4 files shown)


chanind (commented):
I'm hesitant to merge this PR, since this is a breaking change that would cause existing configs to OOM. I agree the config parameter name is confusing, it's really `n_sequences_in_buffer` rather than `n_batches_in_buffer`. Could we instead just change the config param name to `n_sequences_in_buffer` instead, deprecate the old `n_batches_in_buffer` (while still allowing it to work backwards compatibly, we can use an @deprecated flag)? That's probably more clear anyway since `batches` is already a confusing term given there's the LLM batches vs SAE batches, and it's confusing which is which.

danra:
Makes sense to me.

One argument against: perhaps some research was done given an incorrect assumption on how many batches were in the buffer and mixed together. But I suppose making the deprecation and documentation clear on this is good enough; better than unexpectedly OOM-ing, and probably also better than just making a completely breaking change (raising an exception on using old parameter name, not just a deprecation warning).

> That's probably more clear anyway since batches is already a confusing term given there's the LLM batches vs SAE batches, and it's confusing which is which.

Agreed, it confused me for sure.

chanind:
> One argument against: perhaps some research was done given an incorrect assumption on how many batches were in the buffer and mixed together. But I suppose making the deprecation and documentation clear on this is good enough; better than unexpectedly OOM-ing, and probably also better than just making a completely breaking change (raising an exception on using old parameter name, not just a deprecation warning).

If there was research done with an incorrect asumption on the batches in the mixing buffer there's not much we can do about that at this point IMO. I think with a deprecation warning we can output a helpful warning explaining everything to the user and what the issue is as well the new parameter name that's probably the gentlest way we can fix things. If you feel strongly that we should error and crash the run for the user rather than just outputting deprecation warning with a helpful message I'm open to that too, as long as it's extremely clear to the user what to do to fix things. But I would argue for the warning rather than error though, since I imagine a lot users just want to train an SAE and don't care about parameter names and likely don't know what the mixing buffer is to begin with.

danra:
The soft deprecation warning seems like the best solution to me, too.

PR #694, opened 2026-06-07
SAE overrides tidy

danra:
Add overrides decorators consistently in SAE/SAE config classes.
Also fix overwrite->override in a couple of docstrings.


chanind (approved):
👍 Good call!

PR #695, opened 2026-06-07
Add missing calls to reshape_fn_out() in TemporalSAE

danra:
## Type of change

# Checklist:

### DIFF
--- a/sae_lens/saes/temporal_sae.py
+++ b/sae_lens/saes/temporal_sae.py
@@ -342,6 +342,8 @@ class TemporalSAE(SAE[TemporalSAEConfig]):
         # Apply output activation normalization (reverses input normalization)
         sae_out = self.run_time_activation_norm_fn_out(sae_out)
 
+        sae_out = self.reshape_fn_out(sae_out, self.d_head)
+
         logger.warning(
             "NOTE this only decodes x_novel. The x_pred is missing, so we're not reconstructing the full x."
         )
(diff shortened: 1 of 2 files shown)


chanind:
Verified this fix is correct and rebased the branch onto current main (resolving conflicts with #703's conditional `b_dec` and the tests added alongside it — both sets of tests are kept).

Verification notes:

- The added `reshape_fn_out` calls sit at exactly the position every other SAE uses (`StandardSAE.decode`, base `TrainingSAE.decode`): matmul + bias → `hook_sae_recons` → out-normalization → reshape, and in `forward()` before `hook_sae_output`, matching how `_SAEWrapper` fires that hook on reshaped output for other architectures.
- Training is unaffected: `TrainingSAE` turns hook_z reshaping off, and TemporalSAE has no training class.
- End-to-end through `HookedSAETransformer` at `blocks.0.attn.hook_z` on tiny-stories-1M: without this fix, `use_error_term=True` crashes (`original_output - sae_out_clean` broadcasts 4D against flat 3D); with the fix it runs correctly. Interestingly the plain no-error-term path happened to work without the fix because TransformerLens re-flattens hook_z output with a plain reshape — which is why this went unnoticed.
- All 30 temporal tests pass on the rebased branch (28 from main + the 2 added here).

One small change during the rebase: dropped the stale "Add bias (already removed in process_sae_in)" comment the diff carried into `decode()` — the bias is added earlier in the function and main already removed that comment in #702.

chanind:
Thank you for fixing this!

PR #697, opened 2026-06-07
Add covariance whitening normalization option (#647)

azrabano23:
## Summary

Implements #647 — adds **`covariance_whitening`** as a `normalize_activations`
option, applying PCA whitening to activations before the SAE, following
[*Data Whitening Improves Sparse Autoencoder Learning*](https://arxiv.org/abs/2511.13981).

Whitening equalizes the eigenspectrum of the activation covariance, which
de-correlates features and (per the paper) better aligns sparsity with
interpretability.

## How it works

Unlike the other `normalize_activations` options, whitening needs
**data-estimated statistics**, so the SAE stores three buffers:

- `whitening_mean` — activation mean `μ`
- `whitening_W` — whitening matrix `W = D^(-1/2) Eᵀ`
- `whitening_W_inv` — dewhitening matrix `W⁻¹ = E D^(1/2)`

where the covariance `C = E D Eᵀ` and eigenvalues are regularized as `λ + ε`.

The run-time transforms whiten the input and dewhiten the output before the
reconstruction loss (exactly as the paper specifies):

```
in:  z = W (x − μ)
out: x̂ = W⁻¹ z + μ
```

`SAE.estimate_whitening(data_provider, n_batches, eps)` accumulates the mean and
covariance (in float64 for numerical stability), eigendecomposes via
`torch.linalg.eigh`, and fills the buffers. `SAETrainer.fit()` calls it before
training, mirroring the existing `expected_average_only_in` estimation hook. The
buffers live in the state dict, so whitening stays active across save/load
(they are intentionally **not** folded into the weights, since dewhitening is
explicit).

## Tests

`tests/saes/test_sae.py` (functional, not superficial):

- whitened activations have **identity covariance** on correlated synthetic data
- **exact dewhitening round-trip** `out(in(x)) ≈ x`
- whitening is a **no-op before estimation** (buffers start as identity/zero)
- whitening buffers **persist through save/load** (`save_model` → `load_from_disk`)
- `estimate_whitening` raises if the config isn't `covariance_whitening`

`ruff check`, `ruff format`, and `pyright` all pass on the changed files.

## Notes / open questions for maintainers

- Estimation reuses `n_batches_for_norm_estimate` for the number of covariance
  batches. Happy to add a dedicated config field (e.g. `n_batches_for_whitening`)
  if you'd prefer.
- `eps` defaults to `1e-3` (paper uses a small stabilizer); easy to surface as a
  config option if desired.


sae_lens/saes/sae.py
```diff
@@ -353,10 +360,95 @@ def run_time_activation_ln_out(
...
+
+            self.run_time_activation_norm_fn_in = run_time_activation_whiten_in
+            self.run_time_activation_norm_fn_out = run_time_activation_whiten_out
+
         else:
             self.run_time_activation_norm_fn_in = lambda x: x
             self.run_time_activation_norm_fn_out = lambda x: x
 
+    @torch.no_grad()
+    def estimate_whitening(
```
chanind: This makes more sense to go in ActivationScaler, since this whitening is more of a pre-processing of the activations rather than something the base SAE class should need to deal with.

sae_lens/saes/sae.py
```diff
@@ -353,10 +360,95 @@ def run_time_activation_ln_out(
 
             self.run_time_activation_norm_fn_in = run_time_activation_ln_in
             self.run_time_activation_norm_fn_out = run_time_activation_ln_out
+
+        elif self.cfg.normalize_activations == "covariance_whitening":
```
chanind: Can we fold the whitening into the SAE weights at the end of training, like we do with `expected_average_only_in` scaling? This is just a linear operation, so it should be possible to calculate adjusted versions of the weights that take this into account I think. Then we don't need to handle any special processing during inference, only during training, and the distributed SAE doesn't need extra weights matrices.

chanind (commented):
Thank you for taking this task on! The core whitening / unwhitening logic looks good, but the location in the codebase could be improved to better fit with how SAELens training/inference works:

- This looks like an enhancement to the `ActivationScaler` class rather than something that needs to be managed by the base SAE class itself.
- This should only be needed at training time. I think it should be possible to fold whitening into weights when we finish training (see: https://github.com/decoderesearch/SAELens/blob/main/sae_lens/training/sae_trainer.py#L190-L194). We'd need a `fold_` method added to do this covariance weight folding, but then we don't need to save out any new weights for the SAE, and when users load an SAE that was trained with whitening there's no extra compute penalty and no need for special logic in the base SAE class.

azrabano23:
Thanks for the review, and sorry for the delay. I reworked the branch along the lines you suggested and force-pushed (the old commit was conflicting with `main` and almost nothing of it survived the move, so a single fresh commit on top of `main` seemed cleaner than a rebase; your two inline comments pointed at code that no longer exists in `sae.py`).

What changed:

- **Whitening now lives in `ActivationScaler`.** `ActivationScaler.estimate_whitening(d_in, data_provider, n_batches_for_norm_estimate, eps=1e-3)` accumulates the mean and covariance in float64 and stores an `ActivationWhitening(mean, matrix, inverse_matrix)` with `M = E diag(λ+ε)^-1/2`, `M⁻¹ = diag(λ+ε)^1/2 Eᵀ`, so `scale(x) = (x − μ) M` and `unscale(z) = z M⁻¹ + μ`. Evals and the trainer already go through `scale`/`unscale`, so nothing else needed to change there. Checkpoints write the matrices to a sibling `activation_whitening.safetensors` (a `d_in × d_in` matrix did not belong in `activation_scaler.json`); `load` picks it up if present.
- **The base `SAE` class knows nothing about whitening anymore.** No buffers, no runtime branch, no extra weights in saved SAEs. The only addition is `SAE.fold_activation_whitening(mean, whitening_matrix, unwhitening_matrix)`:
  - `W_enc ← M W_enc`
  - `b_enc ← b_enc − μ (M W_enc)`, only when `apply_b_dec_to_input=False` (with it on, the folded `b_dec` already removes `μ` from the input)
  - `W_dec ← W_dec M⁻¹`
  - `b_dec ← b_dec M⁻¹ + μ`
  - Gated applies the same shift to `b_gate` and to `b_mag` (scaled by `exp(r_mag)`, since the magnitude path uses `W_enc · exp(r_mag)`).
  - TopK with `rescale_acts_by_decoder_norm=True` (the training default) folds decoder norms first and clears the flag, because unwhitening changes decoder row norms and would otherwise change the top-k selection; after the norm fold the flag is a no-op, so the result is exact. The saved config therefore ends up with `rescale_acts_by_decoder_norm=False` and unit-norm-then-unwhitened decoder rows. Happy to raise instead if you would rather not have a fold touch that flag.
  - Matching pursuit (tied encoder) and Temporal raise `NotImplementedError`, matching their existing `fold_W_dec_norm` / `fold_activation_norm_scaling_factor` behaviour.
- **Trainer:** `SAETrainer.fit` (and `MultiSAETrainer`) estimate whitening at the start next to the `expected_average_only_in` branch and fold it at the end right after the scaling-factor fold, then clear it, so the final checkpoint has `normalize_activations="none"` and no whitening file.

Tests (`tests/training/test_activation_scaler.py`, `tests/saes/test_sae.py`, `tests/saes/test_topk_sae.py`, `tests/training/test_sae_trainer.py`, `tests/training/test_multi_sae_trainer.py`):

- whitening a fixed sample gives exactly zero mean and identity covariance (atol 1e-4), and `M M⁻¹ = I` to 1e-10
- for every training architecture and both `apply_b_dec_to_input` settings, `sae.encode`/`sae(x)` on raw activations after folding equal the whitened-path features/outputs at `assert_close` defaults in float64 (float64 so a JumpReLU/TopK decision cannot flip from rounding); same for the inference classes, with matching pursuit / temporal asserting the `NotImplementedError`
- a float32 standard SAE agrees to 1e-5 of the output scale (float32 rounding is set by the largest reconstruction, not per element)
- TopK with `rescale_acts_by_decoder_norm=True` folds exactly and the flag is cleared
- end-to-end `SAETrainer.fit` with `normalize_activations="covariance_whitening"`: intermediate checkpoints carry `activation_whitening.safetensors`, the final one does not, and its `cfg.json` says `"none"`; same assertions for `MultiSAETrainer`

`ruff check`, `ruff format --check`, and `pyright` (strict; the only report is the pre-existing optional `mamba_lens` import) pass. Locally (CPU only, Python 3.12, torch 2.14): `pytest tests/saes tests/training/test_activation_scaler.py` → 379 passed; `pytest tests/training/test_sae_trainer.py` → 23 passed, 1 skipped; `pytest tests/training/test_multi_sae_trainer.py` → 12 passed. I have not run `tests/test_evals.py` or the full suite locally (CPU-only laptop), so CI is the check there.

One numerical note in case it matters to you: the whitening tensors are kept in float64 rather than float32. With float32 storage, `M⁻¹ M` deviates from `I` by ~1e-7, which shows up in the `apply_b_dec_to_input=True` fold (`b_dec M⁻¹ M W_enc`) and made the float64 equivalence test fail at ~2e-7; float64 storage makes the fold exact and costs `2·d_in²` doubles at training time only.

PR #698, opened 2026-06-07
chore: layer_norm cleanup similar to the one for constant_norm_rescale

danra:
# Checklist:


chanind (commented):
This looks reasonable, but could we add a test on this as well so we don't regress?

danra:
> This looks reasonable, but could we add a test on this as well so we don't regress?

Done

chanind (approved):
Thank you for this!

PR #700, opened 2026-06-09
Training CLI fixes

### LINKED ISSUE #699: [Bug Report] Training CLI help is wrong for many arguments
Tested on ef7fcacfa210dc53834d57cb2a62b9267962d28d

Training CLI help is wrong for many arguments. For example:

```
-> python sae_lens/llm_sae_training_runner.py --help

...
--device str          Misc (default: cpu)
...
--dtype str           type: ignore # (default: float32)
...
  --reshape_activations {none,hook_z}
                        (Anthropic April 2024 Update) (default: none)
```

This is due to `simple-parsing` favoring comments over docstrings for automatically generating the help strings. This could be easily fixed by adding an option for ignoring comments in`simple-parsing` (preferably upstream, otherwise can fork/vendor).

### Checklist

- [x] I have checked that there is no similar [issue](https://github.com/decoderesearch/SAELens/issues) in the repo (**required**)

danra:
# Description

- Add support for missing architectures in training CLI, list was hard-coded and out-of-date
- Fixes to training CLI help, fixes #699
- As part of the help fixes (adding missing --architecture flag), the CLI code was also simplified by using the existing subgroups feature of simple-parsing

## Type of change

# Checklist:

### DIFF
--- a/docs/training_saes.md
+++ b/docs/training_saes.md
@@ -426,10 +426,10 @@ Resume by passing `resume_from_checkpoint=<checkpoint_dir>`. The per-SAE keys in
 
 ## CLI Runner
 
-The SAE training runner can also be run from the command line via the `sae_lens.sae_training_runner` module. This can be useful for quickly testing different hyperparameters or running training on a remote server. The command line interface is shown below. All options to the CLI are the same as the [LanguageModelSAERunnerConfig][sae_lens.LanguageModelSAERunnerConfig] with a `--` prefix. E.g., `--model_name` is the same as `model_name` in the config.
+The SAE training runner can also be run from the command line via the `sae_lens.llm_sae_training_runner` module. This can be useful for quickly testing different hyperparameters or running training on a remote server. The command line interface is shown below. All options to the CLI are the same as the [LanguageModelSAERunnerConfig][sae_lens.LanguageModelSAERunnerConfig] with a `--` prefix. E.g., `--model_name` is the same as `model_name` in the config.
 
 ```bash
-python -m sae_lens.sae_training_runner --help
+python -m sae_lens.llm_sae_training_runner --help
 ```
 
 ## Logging to Weights and Biases
--- a/sae_lens/saes/batchtopk_sae.py
+++ b/sae_lens/saes/batchtopk_sae.py
@@ -47,25 +47,22 @@ class BatchTopKTrainingSAEConfig(TopKTrainingSAEConfig):
         k (float): Average number of features to keep active across the batch. Unlike
             standard TopK SAEs where k is an integer per sample, this is a float
             representing the average number of active features across all samples in
-            the batch. Defaults to 100.
+            the batch.
         topk_threshold_lr (float): Learning rate for updating the global topk threshold.
             The threshold is updated using an exponential moving average of the minimum
-            positive activation value. Defaults to 0.01.
+            positive activation value.
         aux_loss_coefficient (float): Coefficient for the auxiliary loss that encourages
             dead neurons to learn useful features. Inherited from TopKTrainingSAEConfig.
-            Defaults to 1.0.
         rescale_acts_by_decoder_norm (bool): Treat the decoder as if it was already normalized.
-            Inherited from TopKTrainingSAEConfig. Defaults to True.
+            Inherited from TopKTrainingSAEConfig.
         decoder_init_norm (float | None): Norm to initialize decoder weights to.
-            Inherited from TrainingSAEConfig. Defaults to 0.1.
+            Inherited from TrainingSAEConfig.
         d_in (int): Input dimension (dimensionality of the activations being encoded).
             Inherited from SAEConfig.
         d_sae (int): SAE latent dimension (number of features in the SAE).
             Inherited from SAEConfig.
         dtype (str): Data type for the SAE parameters. Inherited from SAEConfig.
-            Defaults to "float32".
         device (str): Device to place the SAE on. Inherited from SAEConfig.
-            Defaults to "cpu".
     """
 
     k: float = 100  # type: ignore[assignment]
--- a/sae_lens/saes/jumprelu_sae.py
+++ b/sae_lens/saes/jumprelu_sae.py
@@ -193,8 +193,8 @@ class JumpReLUTrainingSAEConfig(TrainingSAEConfig):
         jumprelu_sparsity_loss_mode: mode for the sparsity loss, either "step" or "tanh". "step" is Google Deepmind's L0 loss, "tanh" is Anthropic's sparsity loss.
         l0_coefficient: coefficient for the l0 sparsity loss
         l0_warm_up_steps: number of warm-up steps for the l0 sparsity loss
-        pre_act_loss_coefficient: coefficient for the pre-activation loss. Set to None to disable. Set to 3e-6 to match Anthropic's setup. Default is None.
-        jumprelu_tanh_scale: scale for the tanh sparsity loss. Only relevant for "tanh" sparsity loss mode. Default is 4.0.
+        pre_act_loss_coefficient: coefficient for the pre-activation loss. Set to None to disable. Set to 3e-6 to match Anthropic's setup.
+        jumprelu_tanh_scale: scale for the tanh sparsity loss. Only relevant for "tanh" sparsity loss mode.
     """
 
     jumprelu_init_threshold: float = 0.01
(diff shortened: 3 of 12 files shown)


sae_lens/saes/batchtopk_sae.py
```diff
@@ -47,25 +47,10 @@ class BatchTopKTrainingSAEConfig(TopKTrainingSAEConfig):
...
-            Inherited from TrainingSAEConfig. Defaults to 0.1.
-        d_in (int): Input dimension (dimensionality of the activations being encoded).
-            Inherited from SAEConfig.
-        d_sae (int): SAE latent dimension (number of features in the SAE).
-            Inherited from SAEConfig.
-        dtype (str): Data type for the SAE parameters. Inherited from SAEConfig.
-            Defaults to "float32".
-        device (str): Device to place the SAE on. Inherited from SAEConfig.
-            Defaults to "cpu".
+            positive activation value.
```
chanind, line 53: Why are these other options being deleted from the docs? They should be inherited from the parent class, they are real options that exist

sae_lens/saes/jumprelu_sae.py
```diff
@@ -184,13 +184,14 @@ class JumpReLUTrainingSAEConfig(TrainingSAEConfig):
...
     Configuration class for training a JumpReLUTrainingSAE.
 
-    - jumprelu_init_threshold: initial threshold for the JumpReLU activation
-    - jumprelu_bandwidth: bandwidth for the JumpReLU activation
-    - jumprelu_sparsity_loss_mode: mode for the sparsity loss, either "step" or "tanh". "step" is Google Deepmind's L0 loss, "tanh" is Anthropic's sparsity loss.
-    - l0_coefficient: coefficient for the l0 sparsity loss
-    - l0_warm_up_steps: number of warm-up steps for the l0 sparsity loss
-    - pre_act_loss_coefficient: coefficient for the pre-activation loss. Set to None to disable. Set to 3e-6 to match Anthropic's setup. Default is None.
-    - jumprelu_tanh_scale: scale for the tanh sparsity loss. Only relevant for "tanh" sparsity loss mode. Default is 4.0.
+    Args:
```
chanind, line 190: :+1:

tests/helpers.py
```diff
@@ -36,15 +36,6 @@
...
 ]  # Architectures with fold W_dec to unit norm implementation
-ALL_TRAINING_ARCHITECTURES = [
-    "standard",
-    "gated",
-    "jumprelu",
-    "topk",
-    "batchtopk",
-    "matryoshka_batchtopk",
-    "matching_pursuit",
-]
```
chanind, line 47: nit: could we just replace `ALL_TRAINING_ARCHITECTURES` with `list(SAE_TRAINING_CLASS_REGISTRY.keys())` here? The `ALL_TRAINING_ARCHITECTURES` const feels more understandable than `SAE_TRAINING_CLASS_REGISTRY.keys()` IMO

docs/training_saes.md
```diff
@@ -426,10 +426,10 @@ Resume by passing `resume_from_checkpoint=<checkpoint_dir>`. The per-SAE keys in
 
 ## CLI Runner
 
-The SAE training runner can also be run from the command line via the `sae_lens.sae_training_runner` module. This can be useful for quickly testing different hyperparameters or running training on a remote server. The command line interface is shown below. All options to the CLI are the same as the [LanguageModelSAERunnerConfig][sae_lens.LanguageModelSAERunnerConfig] with a `--` prefix. E.g., `--model_name` is the same as `model_name` in the config.
+The SAE training runner can also be run from the command line via the `sae_lens.llm_sae_training_runner` module. This can be useful for quickly testing different hyperparameters or running training on a remote server. The command line interface is shown below. All options to the CLI are the same as the [LanguageModelSAERunnerConfig][sae_lens.LanguageModelSAERunnerConfig] with a `--` prefix. E.g., `--model_name` is the same as `model_name` in the config.
 
 ```bash
-python -m sae_lens.sae_training_runner --help
+python -m sae_lens.llm_sae_training_runner --help
```
chanind, line 432: 👍

sae_lens/saes/matching_pursuit_sae.py
```diff
@@ -123,31 +104,9 @@ class MatchingPursuitTrainingSAEConfig(TrainingSAEConfig):
     Configuration class for training a MatchingPursuitTrainingSAE.
 
     Args:
-        residual_threshold (float): residual error at which to stop selecting latents. Default 1e-2.
```
chanind, line 126: Why delete the default?

sae_lens/saes/batchtopk_sae.py
```diff
@@ -47,25 +47,10 @@ class BatchTopKTrainingSAEConfig(TopKTrainingSAEConfig):
         k (float): Average number of features to keep active across the batch. Unlike
             standard TopK SAEs where k is an integer per sample, this is a float
             representing the average number of active features across all samples in
-            the batch. Defaults to 100.
```
chanind, line 50: Why remove the default?

tests/test_llm_sae_training_runner.py
chanind, line 855: We should add an assert on the architecutre here, we're currently not testing that it's set correctly

chanind (commented):
This looks like a good improvement, but I'm confused about a few things:

- Why are all the inherited options being removed from the docstring for every SAE class? Are these not needed by mkdocs?
- Why are all default values being removed fro the docstrings? Isn't it important to let users know what the default value is?

One nit on the test refactor, I see the motivation to load the architectures from the registry, but it seems better to keep the named parameter `ALL_TRAINING_ARCHITECTURES` IMO if we can, since it's more readable.

danra:
> This looks like a good improvement, but I'm confused about a few things:
> 
> * Why are all the inherited options being removed from the docstring for every SAE class? Are these not needed by mkdocs?
> * Why are all default values being removed fro the docstrings? Isn't it important to let users know what the default value is?

My rationale for both was DRYness (which wasn't a goal on its own, but it made sense in the context of the CLI arguments now taking their help texts from the docstrings).

Re letting users know about the default values: do you mean via `help()`? If so, the actual default values are shown in the `__init__` signature, that seems visible enough to me to justify removing the duplicate specification of the same values in the docstring.

Re mkdocs: I wasn't familiar with it, looking around a bit it looks like both inherited docstrings (https://github.com/mkdocstrings/griffe-inherited-docstrings) as well as default values should be generated automatically from the code without having to re-specify them in the docstring (unless they are generated dynamically following initialization to `None`, in such cases I kept the docstring default value intact).

> One nit on the test refactor, I see the motivation to load the architectures from the registry, but it seems better to keep the named parameter `ALL_TRAINING_ARCHITECTURES` IMO if we can, since it's more readable.

Sure

danra:
By the way, I recommend reviewing the PR commits separately, it would make the review much clearer

chanind:
> Re mkdocs: I wasn't familiar with it, looking around a bit it looks like both inherited docstrings (https://github.com/mkdocstrings/griffe-inherited-docstrings) as well as default values should be generated automatically from the code without having to re-specify them in the docstring (unless they are generated dynamically following initialization to None, in such cases I kept the docstring default value intact).

It looks like like this requires adding the `griffe-inherited-docstrings` plugin, which is fine, but need to add before merging 

> Re letting users know about the default values: do you mean via help()? If so, the actual default values are shown in the __init__ signature, that seems visible enough to me to justify removing the duplicate specification of the same values in the docstring.

I'm more worried about the SAELens docs, e.g. https://decoderesearch.github.io/SAELens/latest/api/#sae_lens.LanguageModelSAERunnerConfig. It looks like the defaults do broadly show though, so should be fine to delete them

danra:
> It looks like like this requires adding the `griffe-inherited-docstrings` plugin, which is fine, but need to add before merging

I dropped this change from the PR to simplify. (apparently `griffe-inherited-docstrings` is not enough, [docstring-inheritance](https://github.com/AntoineD/docstring-inheritance) (or similar) is also required.)

Other review comments also handled.

chanind:
It seems like the inheritance does work when I run this PR for docs, so maybe none of that was required! Sorry about that, and thank you for fixing this stuff!

danra:
> It seems like the inheritance does work when I run this PR for docs, so maybe none of that was required!

The merged PR left the dup docstrings in place - I dropped de-dupping them. So that's why you still see them correctly :)

chanind:
Aah right, makes sense! We can add those helpers later, seems like a good thing to do regardless. Sorry for the delay reviewing all these PRs, I'll try to get through the rest this weekend!

PR #702, opened 2026-06-11
doc: Fix docstrings/comment in TemporalSAE

danra:


chanind (approved):
Good catch!

PR #706, opened 2026-06-16
Relax tolerance in Qwen Scope loader test to fix flaky CI

### LINKED ISSUE #688: Random test failure
test_qwen_scope_sae_huggingface_loader_with_mocked_download randomly failed in CI (the PR #687 content is likely unrelated) https://github.com/decoderesearch/SAELens/actions/runs/27042996849/job/79822998134.

Possibly the tolerance needs to be relaxed. The max relative error is large, but the assertion doesn't show the actual values, and unfortunately it looks like currently pytest's seed isn't printed out in CI so it's not easily reproducible to examine the actual values.

```
=================================== FAILURES ===================================
_________ test_qwen_scope_sae_huggingface_loader_with_mocked_download __________

tmp_path = PosixPath('/tmp/pytest-of-runner/pytest-0/test_qwen_scope_sae_huggingfac0')
monkeypatch = <_pytest.monkeypatch.MonkeyPatch object at 0x7f1b973a0340>

    def test_qwen_scope_sae_huggingface_loader_with_mocked_download(
        tmp_path: Path, monkeypatch: pytest.MonkeyPatch
    ):
        d_in = 2048
        d_sae = 32 * 1024
    
        # Qwen Scope stores W_enc as (d_sae, d_in) and W_dec as (d_in, d_sae).
        # The loader transposes both to SAELens convention.
        raw_W_enc = torch.randn(d_sae, d_in)
        raw_W_dec = torch.randn(d_in, d_sae)
            11.6723],
         [  -2.7288,   34.1195,  -67.8960,  ...,   16.9614,  -35.8170,
           -54.5632]]])
expected = tensor([[[ -15.0251,  -59.9843,  -26.1955,  ...,  -35.3796,  -26.8852,
            29.8179],
         [  28.6777,   64...7,
            11.6723],
         [  -2.7288,   34.1195,  -67.8960,  ...,   16.9614,  -35.8170,
           -54.5632]]])

    def assert_close(
        actual: torch.Tensor,
        expected: torch.Tensor,
        *,
        allow_subclasses: bool = True,
        atol: float | None = 1e-8,
        rtol: float | None = 1e-5,
        equal_nan: bool = False,
        check_device: bool = True,
        check_dtype: bool = True,
        check_layout: bool = True,
        check_stride: bool = False,
        msg: str | None = None,
    ) -> None:
        """
        torch.testing.assert_close() with torch.allclose() defaults (atol=1e-8, rtol=1e-5).
    
        Pass a message string to customize the error header instead of writing lambda functions.
        """
        final_msg = msg and (lambda error_msg: f"{msg}\n\n{error_msg}")
    
>       torch.testing.assert_close(
            actual,
            expected=expected,
            allow_subclasses=allow_subclasses,
            atol=atol,
            rtol=rtol,
            equal_nan=equal_nan,
            check_device=check_device,
            check_dtype=check_dtype,
            check_layout=check_layout,
            check_stride=check_stride,
            msg=final_msg,
        )
E       AssertionError: Tensor-likes are not close!
E       
E       Mismatched elements: 10288 / 458752 (2.2%)
E       Greatest absolute difference: 0.0001125335693359375 at index (0, 4, 20308) (up to 1e-08 allowed)
E       Greatest relative difference: 0.03291536122560501 at index (1, 5, 26494) (up to 1e-05 allowed)

tests/helpers.py:672: AssertionError
```

robbiebusinessacc:
# Description

  `test_qwen_scope_sae_huggingface_loader_with_mocked_download` intermittently
  fails in CI (e.g. the two runs linked in #688). It compares two mathematically
  equal matmuls: the loader stores `W_enc`/`W_dec` as `.T.contiguous()`, while the
  reference in the test multiplies a transposed view (`raw_W_enc.T`). Over the
  2048- and 32768-dim reductions, a contiguous tensor and a transposed view can
  accumulate float32 in a different order depending on the BLAS backend, so the
  two results agree only up to rounding. The default `assert_close` tolerance
  (`rtol=1e-5`, `atol=1e-8`) is too tight to absorb that, which is what `danra`
  diagnosed on the issue.
  
  This compares with `rtol=1e-3, atol=1e-2`, matching the existing precedent for a
  matmul comparison elsewhere in this file. The tolerance is generous enough to
  absorb the accumulation noise but still catches a real loader regression, which
  would be orders of magnitude off (verified locally: a 1% weight-scale error and
  a wrong-bias error are both still rejected).
  
  Note: the failure is BLAS/threading-dependent and reproduces on CI but not on
  local arm64 macOS (where the two layouts produce bit-identical results), so this
  is verified by the linked CI logs plus the regression check above rather than a
  local repro.
  
  Fixes #688
  
  ## Type of change
  
  # Checklist:

  - [x] I have run the formatter and linter (ruff format + ruff check) on the change.

### DIFF
--- a/tests/loading/test_pretrained_sae_loaders.py
+++ b/tests/loading/test_pretrained_sae_loaders.py
@@ -2364,16 +2364,25 @@ def test_qwen_scope_sae_huggingface_loader_with_mocked_download(
     # The transpose must preserve the underlying linear operation: in the
     # native Qwen Scope formula, pre_acts = residual @ raw_W_enc.T + b_enc.
     # In SAELens, pre_acts = residual @ W_enc + b_enc. They must agree.
+    #
+    # The loader stores W_enc/W_dec as `.T.contiguous()`, so each matmul below
+    # multiplies a contiguous tensor while the reference multiplies a transposed
+    # view. Over these large reduction dims (2048 and 32768) the two layouts can
+    # accumulate float32 in a different order on some BLAS backends, so the
+    # results agree only up to rounding. Use tolerances that absorb that noise
+    # while still catching a real loader regression (which would be orders of
+    # magnitude off); the default rtol=1e-5/atol=1e-8 is too tight here and
+    # causes intermittent CI failures.
     residual = torch.randn(2, 7, d_in)
     expected_pre_acts = residual @ raw_W_enc.T + raw_b_enc
     actual_pre_acts = residual @ state_dict["W_enc"] + state_dict["b_enc"]
-    assert_close(actual_pre_acts, expected_pre_acts)
+    assert_close(actual_pre_acts, expected_pre_acts, rtol=1e-3, atol=1e-2)
 
     # Same for the decoder: native is feats @ raw_W_dec.T + b_dec, SAELens is feats @ W_dec + b_dec.
     feats = torch.randn(2, 7, d_sae)
     expected_recon = feats @ raw_W_dec.T + raw_b_dec
     actual_recon = feats @ state_dict["W_dec"] + state_dict["b_dec"]
-    assert_close(actual_recon, expected_recon)
+    assert_close(actual_recon, expected_recon, rtol=1e-3, atol=1e-2)
 
 
 def test_qwen_scope_sae_loads_via_from_pretrained(


chanind (approved):
Looks good, thanks for this!

PR #708, opened 2026-06-24
feat: add exclude_special_tokens to HookedSAETransformer

### LINKED ISSUE #350: [Proposal] Allow excluding special tokens when applying SAEs in HookedSAETransformer
### Proposal 

We should add an option to exclude special tokens when adding a SAE into `HookedSAETransformer`. This could take the form of an `exclude_special_tokens` param for `add_sae()` / `run_with_cache_with_saes()` / `run_with_saes()`. This would exclude running the SAE on BOS, EOS, and SEP tokens as specified by the model tokenizer. The user could pass `True` to avoid these standard tokens, or pass a list (or tensor) of `token_id` values to exclude to further customize this behavior.

### Motivation

It's often not useful to apply a SAE on special tokens since the SAEs are often not trained on special tokens, and it's not particularly interesting to see SAE latents that fire on BOS. Given this is a common use-case, we should make it easy to just skip special tokens when running with a SAE using `HookedSAETransformer` as this class only exists to make common use-cases for SAEs easy.

### Alternatives

We could alternatively allow users to specify certain token indices to avoid running the SAE on instead of token ids. This would require more work for the users but may support other use-cases where the user doesn't want to apply the SAE at certain positions. This could also be implemented separately / in-addition to adding an `exclude_special_tokens` param.

### Checklist

chandrudp29:
## Summary

- Adds `exclude_special_tokens: bool | list[int] = False` to `add_sae()`, `run_with_saes()`, `run_with_cache_with_saes()`, `run_with_hooks_with_saes()`, and the `saes()` context manager
- When `True`, bypasses SAE processing at BOS/EOS/PAD/SEP/CLS positions using token IDs from the model tokenizer; pass a list of ints for custom token IDs
- Excluded positions return the original activation unchanged
- `exclude_special_tokens` is preserved when `saes()` context manager restores a prior wrapper

## Implementation

`_SAEWrapper` stores a `_token_mask` set per-forward-pass via a `HookedSAETransformer.forward()` override, consumed in `_SAEWrapper.forward()` via `torch.where`. Propagates through all public entry points and preserved across context manager nesting.

All 145 existing tests continue to pass + 10 new tests added.

Closes #350

### DIFF
--- a/sae_lens/analysis/hooked_sae_transformer.py
+++ b/sae_lens/analysis/hooked_sae_transformer.py
@@ -11,6 +11,7 @@ from transformer_lens.HookedTransformer import HookedTransformer
 
 from sae_lens import logger
 from sae_lens.saes.sae import SAE, _disable_hooks
+from sae_lens.util import get_special_token_ids
 
 SingleLoss = torch.Tensor  # Type alias for a single element tensor
 LossPerToken = torch.Tensor
@@ -72,8 +73,8 @@ class _SAEWrapper(nn.Module):
         """
         self._captured_input = x
 
-    def set_token_mask(self, mask: torch.Tensor) -> None:
-        """Set a boolean mask [batch, seq] where True = bypass SAE at that position."""
+    def set_token_mask(self, mask: torch.Tensor | None) -> None:
+        """Set a boolean mask (batch, seq) where True = bypass SAE at that position."""
         self._token_mask = mask
 
     def forward(self, original_output: torch.Tensor) -> torch.Tensor:
@@ -85,6 +86,11 @@ class _SAEWrapper(nn.Module):
             if self._captured_input is not None
             else original_output
         )
+
+        mask = self._token_mask
+        if mask is not None:
+            for _ in range(original_output.dim() - mask.dim()):
+                mask = mask.unsqueeze(-1)
 
         try:
             # Call encode and decode directly so we can control hook_sae_output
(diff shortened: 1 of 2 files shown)


sae_lens/analysis/hooked_sae_transformer.py
```diff
@@ -89,9 +101,18 @@ def forward(self, original_output: torch.Tensor) -> torch.Tensor:
                     sae_error = self.hook_sae_error(original_output - sae_out_clean)
```
chanind: It looks like the sae_error will stil be non-zero on positions where we're ignoring tokens, probably nbd though

sae_lens/analysis/hooked_sae_transformer.py
```diff
@@ -284,12 +322,74 @@ def reset_saes(
 
         self.setup()
 
+    def _get_special_token_ids(self) -> list[int]:
```
chanind: We already have a helper that does this, you can call `get_special_token_ids(tokenizer)` from `sae_lens.util`

sae_lens/analysis/hooked_sae_transformer.py
```diff
@@ -284,12 +322,74 @@ def reset_saes(
...
+                prepend_bos=prepend_bos,
+                padding_side=padding_side,
+                move_to_device=True,
+                truncate=True,
+            )
+            for wrapper in wrappers_needing_mask:
+                mask = self._compute_exclusion_mask(
+                    tokens, wrapper.exclude_special_tokens
+                )
+                wrapper.set_token_mask(mask)
```
chanind: nit: I think this would be better handled as a context manager, so we can always know the token mask is cleaned up at the end

Issue #426, opened 2025-02-11
[Proposal] Document loading existing SAEs for fine-tuning


chanind:
### Proposal 

Our docs don't currently cover this process. We should document how to load an existing SAE and fine-tune it.

meowmaryexe:
Hi! I’m interested in contributing to this issue. My understanding is that SAELens currently doesn’t document the workflow for loading an existing SAE and fine-tuning it. I’d be interested in helping add documentation and examples for this process.

PR #709, opened 2026-06-24
Document fine tuning existing SAEs

J0YY:
Summary
- Document from_pretrained_path for starting a fresh run from saved training SAE weights
- Document resume_from_checkpoint for exact continuation with trainer and activation store state
- Include expected checkpoint files and minimal examples

Verification
- /Users/joyyang/Projects/open-source/.venv/saelens/bin/python -m pytest tests/test_util.py -q
- Markdown code fence count is balanced

Addresses #426

### DIFF
--- a/docs/training_saes.md
+++ b/docs/training_saes.md
@@ -111,6 +111,84 @@ sparse_autoencoder.save_inference_model("path/to/save/dir")
 
 As you can see, the training setup provides a large number of options to explore. The full list of options can be found by inspecting the `LanguageModelSAERunnerConfig` class and the specific SAE configuration class you intend to use (e.g., `StandardTrainingSAEConfig`, `TopKTrainingSAEConfig`, etc.).
 
+## Fine tuning an existing SAE
+
+SAELens supports two different ways to load prior training outputs, depending on
+whether you want a new fine-tuning run or an exact continuation of an existing
+run.
+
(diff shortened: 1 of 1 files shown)


chanind (commented):
Thanks for the contribution! The `from_pretrained_path` documentation was genuinely missing and the technical description was accurate — I verified it against the runner code. I found a few structural issues and pushed a commit (9921d5d7) directly to this branch to address them:

1. **Duplication with the existing Checkpoints section**: `resume_from_checkpoint` was already documented in the "Checkpoints" section further down this same page. To keep each option documented in one place, the resume material now lives only in "Checkpoints" (expanded with the details from this PR), and the fine-tuning section links to it.

2. **Section placement**: the new `##` section was inserted between "Basic training setup" and the per-architecture `###` subsections (Training BatchTopk SAEs, etc.), which made all of those subsections nest under "Fine tuning an existing SAE" in the docs TOC. Moved it after the architecture sections.

3. **Incomplete checkpoint file list**: resuming also requires `activation_scaler.json` — `SAETrainer.load_trainer_state` reads it unconditionally and will crash if it's missing. Added it to the list.

4. Minor: the examples used `device=device` without defining `device`; the `from_pretrained_path` example now defines it (and the resume example in Checkpoints stays minimal).

PR #710, opened 2026-06-28
Fix explained_variance to measure variance relative to the mean (#659)

mkzung:
Fixes #659.

## Problem
The `explained_variance` metric introduced in #443 measures variance relative to **zero** rather than relative to the **per-dimension mean**, which inflates it (the issue reports ~0.98 vs the legacy ~0.78). Two bugs in `get_sparsity_and_variance_metrics`:

1. the per-dimension term used `E[x_d^2]` (an extra `.pow(2)`) instead of `E[x_d]`;
2. `torch.cat` collapsed the per-dimension means `[d_model]` into a scalar, so `sum_d E[x_d]^2` was lost.

Together these make `total_variance ~= E[||x||^2]`, i.e. variance about zero rather than about the mean.

## Fix
Compute `E[x_d]` per dimension and aggregate with `torch.stack`, so `total_variance = sum_d (E[x_d^2] - E[x_d]^2)` (the intended `Var(X) = E[X^2] - E[X]^2`). The aggregation is extracted into a small `_compute_explained_variance` helper so it can be unit-tested directly.

## Test
Added a regression test: predicting the per-dimension mean must explain **0** variance (the pre-fix code reported ~1.0), and perfect reconstruction explains 1.0. `ruff check` and `ruff format` pass.

Thanks to @chanind for the detailed diagnosis in #659.

### DIFF
--- a/sae_lens/evals.py
+++ b/sae_lens/evals.py
@@ -381,6 +381,32 @@ def get_downstream_reconstruction_metrics(
     return metrics
 
 
+def _compute_explained_variance(
+    mean_sum_of_squares: list[torch.Tensor],
+    mean_act_per_dimension: list[torch.Tensor],
+    mean_sum_of_resid_squared: list[torch.Tensor],
+) -> float:
+    """Explained variance of the SAE reconstruction, relative to the per-dimension mean.
+
+    Uses the multi-dimensional identity ``Var(X) = E[X^2] - E[X]^2``::
+
+        total_variance = sum_d (E[x_d^2] - E[x_d]^2)
+        explained_variance = 1 - residual_variance / total_variance
+
+    Each argument holds one entry per eval batch (already reduced over the batch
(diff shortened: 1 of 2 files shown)


mkzung:
The only failure in `build (3.10)` is `tests/synthetic/hierarchy/test_modifier.py::test_hierarchy_modifier_large_hierarchy_performance`, a wall-clock assertion (`6.43s`, threshold `2s`) that is unrelated to this change and flakes depending on runner load. It has also failed recently on unrelated PRs (e.g. `feat/exclude-special-tokens`). Linting, formatting, type checking, and the full `tests/test_evals.py` suite, including the new `test_compute_explained_variance_is_relative_to_the_mean`, all pass. Happy to rebase or adjust if you'd like.

chanind:
Thank you for pushing on this. There was a duplicate fix in #665 that I just merged (should have merged it awhile ago). I'm going to close this as the bug should now be fixed. Apologies for the duplicate work.

PR #714, opened 2026-07-04
Preserve training architecture in SAE metadata

abhinav-bellapu:
## Summary

- Add deterministic `training_architecture` and `training_architecture_details` metadata to training SAE configs.
- Preserve that provenance when converting training configs into inference SAE configs, including BatchTopK/MatryoshkaBatchTopK saves that load as JumpReLU inference SAEs.
- Cover config serialization, single-SAE training runner saves, checkpoints, and multi-SAE training saves with regression tests.

## Motivation

Issue #663 points out that saved configs can lose the training architecture once a training-only architecture is represented by its inference architecture. This records the original training architecture and the training-only config fields in metadata so downstream users can inspect how the SAE was trained even when the runtime inference architecture differs.

This PR addresses the training-architecture provenance part of #663. It intentionally does not add final l0 or dead-latent counts to config metadata, since those are run/evaluation artifacts rather than deterministic config fields.

## Compatibility and performance

This is an additive metadata-only change. Existing configs that do not contain these metadata keys remain loadable. It does not change model weights, forward or inference behavior, training losses, optimization, or checkpoint tensors.

Performance metrics (L0, CE loss, MSE loss, and feature-dashboard interpretability) are therefore unchanged / not applicable. Final L0 and dead-latent counts remain excluded because they are run/evaluation outputs rather than deterministic configuration.

## Validation

- `.venv/bin/python -m pytest tests/saes/test_sae.py -q`
- `.venv/bin/python -m pytest tests/training/test_config.py::test_LanguageModelSAERunnerConfig_to_dict_and_from_dict -q`
- `.venv/bin/python -m pytest tests/test_llm_sae_training_runner.py::test_LanguageModelSAETrainingRunner_runs_and_saves_all_architectures tests/test_llm_sae_training_runner.py::test_LanguageModelSAETrainingRunner_saves_final_output_and_checkpoints tests/test_multi_sae_training_runner.py::test_multi_sae_runner_trains_two_saes_at_different_hooks -q`
- `.venv/bin/ruff format --check sae_lens/saes/sae.py tests/saes/test_sae.py tests/test_llm_sae_training_runner.py tests/test_multi_sae_training_runner.py`
- `.venv/bin/ruff check sae_lens/saes/sae.py tests/saes/test_sae.py tests/test_llm_sae_training_runner.py tests/test_multi_sae_training_runner.py`
[...]

PR #717, opened 2026-07-19
feat: record final l0 and num dead features in SAE metadata

shubhangithub:
## Summary

Records final `l0` and `num_dead_features` in the SAE metadata at the end of training, so both end up in the saved `cfg.json` of the final checkpoint and of the exported inference SAE. `MultiSAETrainer` does the same for each of its per-SAE trainers.

This finishes the remaining half of #663: #714 added the training-architecture metadata, and the follow-up comment asked for l0 and num dead latents recorded at the end of training.

Fixes #663

## Implementation notes

- Both values come from the trainer's existing sparsity tracking, no extra per-step work. `l0` is the sum of per-feature firing probabilities over the current sparsity window, i.e. the window-averaged mean l0 (same firing stats as the `metrics/l0` wandb log, which reports the per-batch value). `num_dead_features` matches the `sparsity/dead_features` log. If training never stepped, the fields are left unset rather than storing NaN.
- The metadata is set in `SAETrainer.fit()` just before the final checkpoint save rather than in each runner. That keeps the final checkpoint and the exported inference SAE consistent, and covers the LLM, multi-SAE, and synthetic runners in one place. Intermediate checkpoints don't get these fields since they're end-of-training stats.
- Named the count `num_dead_features` to match `dead_feature_window` and the `sparsity/dead_features` log key. The issue says "dead latents", happy to rename.
- Left out `num_training_samples` since the comment said not to go overboard and `runner_cfg.json` already records it.

## Compatibility

Additive metadata-only change: no training math, weights, or checkpoint tensors change. Configs without the new keys load as before (`SAEMetadata` returns `None` for missing keys). One existing test (`test_checkpoints_save_runner_cfg`) was updated to expect the new keys in the final checkpoint's metadata.

## Validation

- `pytest tests/training/test_sae_trainer.py tests/training/test_multi_sae_trainer.py tests/test_llm_sae_training_runner.py tests/test_multi_sae_training_runner.py` all pass
- `ruff format --check` and `ruff check` on the changed files
- `pyright` on the changed files, no new errors vs `main`

Like #714 I used a uv-managed Python environment locally rather than Poetry.

### DIFF
--- a/sae_lens/training/multi_sae_trainer.py
+++ b/sae_lens/training/multi_sae_trainer.py
@@ -149,6 +149,9 @@ class MultiSAETrainer:
                 )
                 trainer.activation_scaler.scaling_factor = None
 
+        for trainer in self.trainers.values():
+            trainer.set_final_sae_metadata()
+
         if self.cfg.save_final_checkpoint:
             self.save_checkpoint(checkpoint_name=f"final_{self.n_training_samples}")
 
--- a/sae_lens/training/sae_trainer.py
+++ b/sae_lens/training/sae_trainer.py
@@ -193,11 +193,31 @@ class SAETrainer(Generic[T_TRAINING_SAE, T_TRAINING_SAE_CONFIG]):
             )
             self.activation_scaler.scaling_factor = None
 
+        self.set_final_sae_metadata()
+
         if self.cfg.save_final_checkpoint:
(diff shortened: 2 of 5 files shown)


chanind (approved):
This looks like a great addition, thank you for this!

PR #718, opened 2026-07-20
fix: parameterize JumpReLU threshold directly instead of log_threshold

### LINKED ISSUE #494: [Bug Report][alpha branch]JumpReLU SAEs not reducing L0 in training
**Describe the bug**
On the current alpha branch, JumpRELU SAEs fail to reduce L0 in training, regardless of L0 coefficient. See for example this wandb plot:

Image

**Minimal example/context**
Context: I'm experimenting with a novel SAE architecture, implemented as a subclass of `TrainingSAE`, and wanted to generate a baseline comparison using JumpReLU. When I swap the JumpReLU with a StandardSAE in the below setup, I get a reasonably sparse result, so this seems to be JumpReLU-specific.

Adapted from the [tutorial](https://github.com/jbloomAus/SAELens/blob/main/tutorials/training_a_sparse_autoencoder.ipynb), accounting for the refactors in alpha:

```
from sae_lens import (
    JumpReLUTrainingSAEConfig,
    LanguageModelSAERunnerConfig,
    SAETrainingRunner,
    TrainingSAE,
)
from sae_lens.config import LoggingConfig
from sae_lens.saes.sae import SAEMetadata

context_size = 1024
total_training_steps = 2_000
batch_size = 10 * context_size
total_training_tokens = total_training_steps * batch_size
model_name = "gpt2-small"
hook_layer = 5
hook_name = f"blocks.{hook_layer}.hook_resid_post"
dataset_path = "monology/pile-uncopyrighted"
expansion_factor = 4

    eval_batch_size_prompts=16,
    model_name=model_name,
    hook_name=hook_name,
    hook_layer=5,
    sae=JumpReLUTrainingSAEConfig(
        metadata=SAEMetadata(
            model_name=model_name,
            hook_name=hook_name,
            hook_layer=hook_layer,
            context_size=context_size,
            dataset_path=dataset_path,
        ),
        d_sae=768 * expansion_factor,
        normalize_activations="expected_average_only_in",
        d_in=768,
        mse_loss_normalization="dense_batch",
        device="cuda",
        l0_coefficient=100,
        l0_warm_up_steps=total_training_steps // 20,  # 5% of training,
    ),
    dataset_path=dataset_path,
    is_dataset_tokenized=False,
    streaming=True,  # we could pre-download the token dataset if it was small.
    lr=1e-5,  # lower the better, we'll go fairly high to speed up the tutorial.
    adam_beta1=0.9,  # adam params (default, but once upon a time we experimented with these.)
    adam_beta2=0.999,
    lr_scheduler_name="constant",  # constant learning rate with warmup. Could be better schedules out there.
    lr_warm_up_steps=0,  # this can help avoid too many dead features initially.
    lr_decay_steps=total_training_steps // 5,  # 20% of training
    train_batch_size_tokens=batch_size,
    context_size=context_size,  # will control the lenght of the prompts we feed to the model. Larger is better but slower. so for the tutorial we'll use a short one.
    # Activation Store Parameters
    n_batches_in_buffer=64,  # controls how many activations we store / shuffle.
    training_tokens=total_training_tokens,  # 100 million tokens is quite a few, but we want to see good stats. Get a coffee, come back.
    store_batch_size_prompts=32,
    # Resampling protocol
    feature_sampling_window=1000,  # this controls our reporting of feature sparsity stats
    dead_feature_window=1000,  # would effect resampling or ghost grads if we were using it.
    dead_feature_threshold=1e-4,  # would effect resampling or ghost grads if we were using it.
    # WANDB
    logger=LoggingConfig(
        log_to_wandb=True,  # always use wandb unless you are just testing code.
        wandb_project="jumprelu-repro",
        wandb_log_frequency=30,
        eval_every_n_wandb_logs=20,
    ),
    # Misc
    device="cuda",
    seed=42,
    n_checkpoints=0,
    checkpoint_path="/tmp/checkpoints",
    dtype="float32",
)
# Looks like a bug in the alpha branch; the default code path filters out any config
# values not present in the base config class
override_sae = TrainingSAE.from_dict(cfg.get_training_sae_cfg_dict())
runner = SAETrainingRunner(cfg, override_sae=override_sae)

sae = runner.run()
```

Note also: I am using `override_sae` because [this code path](https://github.com/jbloomAus/SAELens/blob/f1c0586cd98efb3c8c7904ef1dc14de5f3343195/sae_lens/sae_training_runner.py#L82-L86) causes architecture-specific config options to be reset to default, due to instantiating a generic config class that then filters unknown options; I suspect this line should just be `TrainingSAE.from_dict(self.cfg.get_training_sae_cfg_dict())`.

**System Info**
Describe the characteristic of your environment:
 * Describe how `transformer_lens` was installed (pip, docker, source, ...)
 `uv` lockfile
 * What OS are you using? (Linux, MacOS, Windows)
 Linux
 * Python version (We suppourt 3.10 -3.12 currently)
3.11.5

### Checklist

Hetens:
# Description

`JumpReLUTrainingSAE` stored its threshold as `log_threshold` and exposed `threshold = exp(log_threshold)`. Under Adam this makes the threshold effectively unlearnable, so the L0 penalty cannot reduce L0 no matter how large `l0_coefficient` is.

The reasoning:

1. In JumpReLU, the sparsity penalty's only gradient path is the threshold. `Step.backward` returns `None` for the `x` gradient and a gradient only for `threshold`, so if the threshold cannot move, L0 cannot change.
2. Adam normalizes by gradient magnitude, so a step moves a parameter by at most about `lr`, regardless of how large the gradient is.
3. With `threshold = exp(log_threshold)`, a step of `~lr` in log space moves the *effective* threshold by only `~threshold * lr`. At the default `jumprelu_init_threshold=0.01` and typical learning rates that is on the order of `1e-7` per step, so the threshold stays pinned at its init value for any realistic run.

This also explains the specific symptom reported in #494 — that raising `l0_coefficient` changes nothing. Adam divides the gradient by its own magnitude, so scaling the loss term scales numerator and denominator together and cancels out.

The fix is to parameterize the threshold directly, matching DeepMind's JumpReLU pseudocode and `saprmarks/dictionary_learning`, which uses a raw threshold and explicitly notes poor results with a log threshold.

Fixes #494

## Notes for reviewers

- **The public checkpoint format does not change.** The old code already converted `log_threshold` to `threshold` in `process_state_dict_for_saving`, so on-disk checkpoints were always keyed `threshold`. Existing saved JumpReLU SAEs load unchanged. `process_state_dict_for_loading` now converts legacy `log_threshold` keys forward, covering any in-memory state dict produced by the old code, and there is a test for this.
- **The diff is small because the abstraction was already right.** Every call site already read `self.threshold`; only the storage changed.
- **Positivity is no longer structurally guaranteed.** The log parameterization enforced `threshold > 0` for free. In the reference implementations, and in the runs below, the threshold is driven upward by the sparsity term and does not go negative in practice, and a negative threshold degrades gracefully to a plain ReLU rather than producing NaNs. Happy to add a clamp if you would prefer the hard guarantee.

## Type of change

# Checklist:

- [ ] I have made corresponding changes to the documentation — not applicable, no docs referenced `log_threshold`

`test_JumpReLUTrainingSAE_threshold_travels_at_the_adam_step_size` is the regression test. I verified it actually catches the bug by running it against unmodified `main`, where it fails at `0.00031` moved against a `0.01` bar; on this branch it passes at `0.031`. The bounds were calibrated over repeated runs (spread `0.0306`–`0.0310`) so it does not need a seed and should not flake.

I could not run `make check-ci` end to end because I do not have the poetry environment set up locally. What I did run: `ruff format`, `ruff check` and `pyright` are all clean on both changed files, and `pytest tests/saes/` passes (260 passed). Two files in `tests/saes/` fail collection in my environment for an unrelated missing optional dependency (`sparsify`), on `main` as well as on this branch. Please let me know if CI surfaces anything and I will fix it.

### Performance Check.

- [x] L0
- [x] MSE Loss
- [ ] CE Loss — not measured
- [ ] Feature Dashboard Interpretability — not measured

Control and test groups are both here, as a wandb project with all six runs overlaid:

**https://wandb.ai/hetansh-shah95-umass-amherst/saelens-494-jumprelu-threshold**

The `threshold_mean` panel is the one to look at: the control arms sit flat at the `0.01` init while the fixed arms climb. Each run's config records `threshold_travel_ceiling = lr * steps`, so the claim that Adam bounds the threshold's movement can be checked directly against the logged curve rather than taken on trust.

Both experiments hold the data, seed, weight init and batch order fixed and vary only the parameterization. The full-pipeline results below were run locally and are reported as tables rather than dashboards.

**Experiment 1 — offline, on cached gpt2 `blocks.5.hook_resid_post` activations.** 200k activations, `d_sae=3072`, 2000 steps. "log" reinstates the current parameterization as a subclass so both arms run identical forward math.

| wandb run | parameterization | lr | l0 coef | final L0 | final threshold | final MSE |
|---|---|---|---|---|---|---|
| `log_lr1e-05_coef100` | log (current) | 1e-5 | 100 | 2863 | 0.0101 | 98.50 |
| `raw_lr1e-05_coef100` | raw (this PR) | 1e-5 | 100 | 2848 | 0.0237 | 98.49 |
| `log_lr0.0003_coef100` | log (current) | 3e-4 | 100 | 3019 | 0.0157 | 0.68 |
| `raw_lr0.0003_coef100` | raw (this PR) | 3e-4 | 100 | **1021** | **0.6197** | 15.06 |
| `raw_lr0.0003_coef10` | raw (this PR) | 3e-4 | 10 | 1021 | 0.6192 | 14.93 |
| `raw_lr0.0003_coef1` | raw (this PR) | 3e-4 | 1 | 1021 | 0.6278 | 15.55 |

The current parameterization leaves the threshold within a few percent of its `0.01` init at *any* learning rate, with L0 stuck near 3000 of 3072 latents, i.e. essentially dense. Raising lr 30x does not rescue it: `0.0101` at 1e-5 becomes only `0.0157` at 3e-4. With the fix the threshold reaches `0.62` and L0 falls to 1021, still decreasing at step 2000 — a 39x difference in threshold against the matched control.

MSE rises from `0.68` to `15.06` in that comparison. That is the expected reconstruction cost of actually enforcing sparsity: the low-MSE run is low-MSE precisely because it is dense and reconstructing from ~3000 active latents.

**Experiment 2 — full training pipeline**, streamed `monology/pile-uncopyrighted` through gpt2, matching the reporter's setup. 2000 steps, `d_sae=3072`, `l0_coefficient=100`, threshold init `0.01`.

| branch | lr | final mean threshold | final L0 |
|---|---|---|---|
| `main` | 1e-5 | 0.0102 | ~2888 |
| this PR | 1e-5 | 0.0264 | ~2638 |
| this PR | 5e-5 | 0.0918 | ~3023 |

## Two caveats I want to state rather than have you find

**1. The fix removes the hard block, but short runs still will not produce sparse SAEs.** Because Adam caps movement at `~lr` per step, the threshold's reachable range is bounded by roughly `lr * steps`. The runs above sit essentially at that ceiling — `5e-5 * 2000 = 0.10` versus `0.0918` reached, and `3e-4 * 2000 = 0.60` versus `0.6197` reached — so the optimizer is already moving the threshold as fast as it can. At `lr=5e-5` for 2000 steps L0 has not dropped yet: the threshold is climbing but the encoder's pre-activations grow under the MSE term at the same time, and the crossover has not happened. In the offline run at `3e-4` the crossover occurs around step 800, after which L0 falls sharply — visible in the `l0` panel of the wandb workspace. So reaching a useful L0 needs an adequate `lr * steps` budget; the tutorial defaults are below it.

**2. `l0_coefficient` has little effect within these runs, even after the fix** — L0 lands at 1021.2 / 1021.0 / 1020.9 for coefficients 100 / 10 / 1, a spread of 0.3 latents across a 100x change in the penalty weight. This is the same Adam normalization at work and I think it is worth knowing about, though it is a separate question from this bug. If you would rather the coefficient behave like a real knob, that likely means excluding the threshold from Adam's normalization — a larger change that I would rather not fold into a fix PR. Happy to open a follow-up issue.

Reproduction scripts for both experiments are available if useful.

### DIFF
--- a/sae_lens/saes/jumprelu_sae.py
+++ b/sae_lens/saes/jumprelu_sae.py
@@ -12,6 +12,7 @@ from sae_lens.saes.sae import (
     TrainingSAE,
     TrainingSAEConfig,
     TrainStepInput,
+    TrainStepOutput,
 )
 
 
@@ -143,7 +144,7 @@ class JumpReLUSAE(SAE[JumpReLUSAEConfig]):
 
         # 2) Zero out any unit whose (hidden_pre <= threshold).
         #    We cast the boolean mask to the same dtype for safe multiplication.
-        jump_relu_mask = (hidden_pre > self.threshold.relu()).to(base_acts.dtype)
+        jump_relu_mask = (hidden_pre > self.threshold).to(base_acts.dtype)
 
         # 3) Multiply the normally activated units by that mask.
         return self.hook_sae_acts_post(base_acts * jump_relu_mask)
@@ -270,6 +271,17 @@ class JumpReLUTrainingSAE(TrainingSAE[JumpReLUTrainingSAEConfig]):
         )
 
     @override
+    def training_forward_pass(self, step_input: TrainStepInput) -> TrainStepOutput:
+        # A directly parameterized threshold can be driven below zero, where the
+        # JumpReLU gate would pass negative pre-activations through as negative
+        # feature activations. Project it back onto [0, inf) instead of clamping
+        # inside the graph: clamping in the graph zeroes the threshold's own
+        # gradient, which pins any latent that reaches zero there permanently.
+        with torch.no_grad():
+            self.threshold.data.clamp_(min=0.0)
+        return super().training_forward_pass(step_input)
+
+    @override
     def encode_with_hidden_pre(
         self, x: torch.Tensor
     ) -> tuple[torch.Tensor, torch.Tensor]:
@@ -277,7 +289,7 @@ class JumpReLUTrainingSAE(TrainingSAE[JumpReLUTrainingSAEConfig]):
 
         hidden_pre = self.hook_sae_acts_pre(sae_in @ self.W_enc + self.b_enc)
         feature_acts = self.hook_sae_acts_post(
-            JumpReLU.apply(hidden_pre, self.threshold.relu(), self.bandwidth)
+            JumpReLU.apply(hidden_pre, self.threshold, self.bandwidth)
         )
 
         return feature_acts, hidden_pre  # type: ignore
@@ -292,9 +304,7 @@ class JumpReLUTrainingSAE(TrainingSAE[JumpReLUTrainingSAEConfig]):
     ) -> dict[str, torch.Tensor]:
         """Calculate architecture-specific auxiliary loss terms."""
 
-        # relu() keeps the threshold >= 0: a negative threshold disables the
-        # gate (see JumpReLUSAE.encode) and counts the latent as always-on in L0.
-        threshold = self.threshold.relu()
+        threshold = self.threshold
         W_dec_norm = self.W_dec.norm(dim=1)
         if self.cfg.jumprelu_sparsity_loss_mode == "step":
             l0 = torch.sum(
--- a/sae_lens/saes/jumprelu_sae.py
+++ b/sae_lens/saes/jumprelu_sae.py
@@ -222,17 +222,23 @@ class JumpReLUTrainingSAE(TrainingSAE[JumpReLUTrainingSAEConfig]):
       - A learnable threshold parameter.
       - A specialized auxiliary loss term for sparsity (L0 or similar).
 
-    The threshold is parameterized directly rather than as exp(log_threshold):
-    under Adam the per-step parameter movement is capped at ~lr regardless of
-    gradient magnitude, so a log parameterization moves the effective threshold
-    by only ~threshold * lr per step. With a small init this freezes the
-    threshold and the L0 penalty cannot reduce L0 at all (issue #494).
+    The threshold is parameterized directly rather than as exp(log_threshold).
+    Under Adam a parameter moves by at most ~lr per step regardless of gradient
+    magnitude, so a log parameterization moves the threshold multiplicatively
+    and a direct one moves it additively: growing a small init to a useful
+    threshold costs lr * steps of about ln(target / init) in log space versus
+    about target here. That gap decides whether the L0 penalty can do anything
+    at the run lengths SAELens is typically used for. Training gpt2-small at
+    lr 1e-5, a log threshold barely left its 0.01 init and L0 stayed at
+    3072/3072, while a direct threshold brought L0 below 1000 (issue #494). At
+    lr 3e-4 both reach a comparable solution, so this fixes the small
+    lr * steps regime rather than a defect present at every scale.
 
-    Note this diverges from DeepMind's JumpReLU (paper appendix J and their
-    colab), which trains log(threshold) to keep the threshold positive.
-    saprmarks/dictionary_learning instead parameterizes the threshold directly
-    and notes poor results with a log threshold, consistent with what we see
-    here.
+    This diverges from DeepMind's JumpReLU (paper appendix J and their colab),
+    which trains log(threshold) partly to keep the threshold positive; that is
+    maintained here by projecting the parameter in training_forward_pass.
+    saprmarks/dictionary_learning also parameterizes the threshold directly and
+    notes poor results with a log threshold.
 
     Methods of interest include:
 
--- a/sae_lens/saes/jumprelu_sae.py
+++ b/sae_lens/saes/jumprelu_sae.py
@@ -1,7 +1,6 @@
 from dataclasses import dataclass
 from typing import Any, Literal
 
-import numpy as np
 import torch
 from torch import nn
 from typing_extensions import override
@@ -219,18 +218,26 @@ class JumpReLUTrainingSAE(TrainingSAE[JumpReLUTrainingSAEConfig]):
 
     Similar to the inference-only JumpReLUSAE, but with:
 
-      - A learnable log-threshold parameter (instead of a raw threshold).
+      - A learnable threshold parameter.
       - A specialized auxiliary loss term for sparsity (L0 or similar).
+
+    The threshold is parameterized directly rather than as exp(log_threshold):
+    under Adam the per-step parameter movement is capped at ~lr regardless of
+    gradient magnitude, so a log parameterization moves the effective threshold
+    by only ~threshold * lr per step. With a small init this freezes the
+    threshold and the L0 penalty cannot reduce L0 at all (issue #494). The
+    reference implementations (DeepMind's pseudocode, dictionary_learning)
+    parameterize the threshold directly.
 
     Methods of interest include:
 
-    - initialize_weights: sets up W_enc, b_enc, W_dec, b_dec, and log_threshold.
+    - initialize_weights: sets up W_enc, b_enc, W_dec, b_dec, and threshold.
     - encode_with_hidden_pre_jumprelu: runs a forward pass for training.
     - training_forward_pass: calculates MSE and auxiliary losses, returning a TrainStepOutput.
     """
 
     b_enc: nn.Parameter
-    log_threshold: nn.Parameter
+    threshold: nn.Parameter
 
     def __init__(self, cfg: JumpReLUTrainingSAEConfig, use_error_term: bool = False):
         super().__init__(cfg, use_error_term)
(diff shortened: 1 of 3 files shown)


chanind:
Thank you for digging into this! The original jumprelu code is from Deepmind's colab here: https://colab.research.google.com/drive/1PlFzI_PWGTN9yCQLuBcSuPJUjgHL7GiD. This does store `log_threshold` like we do here, and the [JumpReLU paper](https://arxiv.org/abs/2407.14435) appendix J does specifically mention using a log threshold:

> We use the logarithm of threshold, i.e. log(𝜽), as our trainable parameter, to ensure that the threshold remains positive during training. 

This change is thus diverging from Deepmind's implementation.  That divergence from the official JumpReLU recipe is fine if it fixes things IMO, and you make a compelling case.

> saprmarks/dictionary_learning, which uses a raw threshold and explicitly notes poor results with a log threshold

Where do you see this? I see they don't use a log threshold for their JumpReLU implementation, but can't find this discussion.

Hetens:
I tried the fix because the comment on lines 148-150 in saprmarks/dictionary_learning says they had poor results with `log_threshold` and switched to a raw threshold: [The comment](https://github.com/saprmarks/dictionary_learning/blob/main/dictionary_learning/trainers/jumprelu.py#L148-L150)

I hypothesize the prime reason is that in log space, under Adam, the threshold step is capped at ~`threshold·lr`, so at the default 0.01 init it moves only ~1e-7/step and basically can't get sparse, while a raw threshold moves at ~`lr`. The controlled runs in the PR line up with this: same seed/data/init, only the parameterization changes, and the log threshold stays pinned near the init (L0 stuck ~3000) while the raw one reaches ~0.62 with L0 dropping.

chanind:
Your reasoning makes sense and it makes sense this could explain the poor results they saw in the dictionary_learning library. I think that if we do this, we should probably add a `relu()` before the jumprelu stuff gets applied though, since otherwise we can get negative thresholds and that's not possible with the current jumprelu inference SAE.

sae_lens/saes/jumprelu_sae.py
```diff
@@ -219,50 +218,53 @@ class JumpReLUTrainingSAE(TrainingSAE[JumpReLUTrainingSAEConfig]):
...
+      - A learnable threshold parameter.
       - A specialized auxiliary loss term for sparsity (L0 or similar).
 
+    The threshold is parameterized directly rather than as exp(log_threshold):
+    under Adam the per-step parameter movement is capped at ~lr regardless of
+    gradient magnitude, so a log parameterization moves the effective threshold
+    by only ~threshold * lr per step. With a small init this freezes the
+    threshold and the L0 penalty cannot reduce L0 at all (issue #494). The
+    reference implementations (DeepMind's pseudocode, dictionary_learning)
+    parameterize the threshold directly.
```
chanind: This comment is not correct, we are doing the opposite of DeepMind's implementation if we do this

Hetens: Yeah, I saw the colab notebook. My bad, nevertheless it led to the solution.

Hetens:
A negative threshold doesn't break inference. In `JumpReLUSAE.encode` the output is `ReLU(hidden_pre) * (hidden_pre > threshold)` ([`jumprelu_sae.py#L142-L149`](https://github.com/Hetens/SAELens/blob/4d5137eb16d52377b909e79f0c72c979da74b83e/sae_lens/saes/jumprelu_sae.py#L142-L149); `activation_fn` is `nn.ReLU`, [`sae.py#L318`](https://github.com/Hetens/SAELens/blob/4d5137eb16d52377b909e79f0c72c979da74b83e/sae_lens/saes/sae.py#L318)). For `threshold < 0`, the mask `(h > threshold)` is `True` everywhere `ReLU(h) > 0`, so the gate stops removing anything and the latent computes `ReLU(h)`.

It's still worth clamping, for two reasons: (1) that latent then fires on every positive pre-activation, so it counts as always-on in the L0 sum and inflates the reported L0; (2) once the threshold is negative there's no pressure pulling it back positive, so the latent stays a plain ReLU instead of a gated JumpReLU. So `relu()`/`clamp(min=0)` on the threshold to keep the L0 count correct and the latent gated.

Issue #719, opened 2026-07-20
[Bug Report] l0_coefficient has almost no effect on JumpReLU sparsity (Adam normalizes it away)


Hetens:
**Describe the bug**

`l0_coefficient` has almost no effect on the sparsity reached by a JumpReLU SAE, across three orders of magnitude. This is separate from #494 (threshold parameterization) and **survives the fix in #718**.

The issue was reported earlier in #382 ("L0 lambda variations (50, 100, 500) show minimal impact on L0 values") but never solved. I'm opening this separately because #718 fixes the frozen-threshold half of #382 and this half remains.

The cause is that the L0 penalty's only gradient path is the threshold, and Adam normalizes by gradient magnitude. Adam's update is `lr * m̂ / √v̂`; scaling the loss by a constant `c` scales `m̂` and `√v̂` equally, so `c` cancels and the step size is unchanged. The coefficient changes the reported loss but not the optimization.

For the reconstruction/sparsity tradeoff this matters more than usual: `l0_coefficient` is the documented knob for choosing where an SAE lands on that tradeoff, and right now turning it does nothing.

**Code example**

Controlled runs on cached gpt2 `blocks.5.hook_resid_post` activations (200k tokens, 2000 steps, `d_sae=3072`, `lr=3e-4`, bandwidth `0.05`), on the #718 branch so the threshold is free to move. Same seed, init and data order; only `l0_coefficient` varies:

| `l0_coefficient` | final L0 | final threshold | final MSE |
|---|---|---|---|
| 100 | 1021.2 | 0.6197 | 15.06 |
| 10 | 1021.0 | 0.6192 | 14.93 |
| 1 | 1021.1 | 0.6278 | 15.55 |

A 100x change in the coefficient moves L0 by 0.03%.

Runs: https://wandb.ai/hetansh-shah95-umass-amherst/saelens-494-jumprelu-threshold

**System Info**

- `SAELens` installed from source (current `main`, plus the branch in #718)
- Linux
- Python 3.11

**Additional context**

What actually controls the final L0 today is `lr * n_steps`, since Adam caps threshold movement at ~`lr` per step and runs end within a few percent of that ceiling. I don't think it obviously stands out at first.

I don't have a fix I'm confident in, so I'd rather raise it than guess. Possible experiments;

[...]

PR #720, opened 2026-07-27
docs: document that l0_coefficient saturates, and warn when it has

guptaishaan:
Addresses #719. Builds on #718 — that PR frees the JumpReLU threshold to move; this is about what happens once it can.

> **Note on the diff:** this branch is stacked on #718, so it carries that PR's four commits as well. Only the last commit (`docs: document that l0_coefficient saturates, and warn when it has`) is new here. It needs #718 because on `main` the threshold is a read-only `exp(log_threshold)` property that barely moves, so the diagnostic below can't apply. Happy to rebase once #718 lands, or to retarget it if #718 changes shape.

#719 reports that `l0_coefficient` does nothing across 1/10/100. It isn't inert, it saturates. Sweeping downward, on cached activations with `lr=3e-4`, bandwidth 0.05, init threshold 0.01, 2000 steps:

| `l0_coefficient` | 0 | 1e-3 | 1e-2 | 0.1 | 1 | 10 | 100 |
|---|---|---|---|---|---|---|---|
| final L0 | 478 | 364 | 124 | 36 | 26.6 | 26.4 | 26.0 |

The knob moves L0 10x over `[1e-3, 1e-1]` and is flat above ~1, so the values in the report were all past the point where it stops doing anything.

The threshold receives `c * g_l0 + g_mse`, and Adam fixes the step size at ~`lr`, so `c` only sets the *direction* of the update. Measured mid-training, the unit-`c` l0 gradient on the threshold is ~1000x the reconstruction gradient, so by `c ≈ 0.01` the direction is already entirely the l0 term and further increases are no-ops.

Where that boundary sits depends on activation scale, `jumprelu_bandwidth`, `d_sae`, batch size and `n_steps` — at 20k steps instead of 2k the same sweep is still responsive at `c=1 → 10`. So a better default wouldn't fix it.

**This PR doesn't change any optimization behaviour.** It:

- documents the saturation, scoped to `step` mode — in `tanh` mode the penalty also back-propagates through `W_enc`/`b_enc`/`W_dec` and `l0_coefficient` responds normally (L0 3.84 → 2.59 → 0.84 across c = 1/10/100), so the "only gradient path is the threshold" reasoning doesn't apply there
- warns at the end of training when the threshold travelled ≥90% of what Adam permits, which is the signal that the run is in the saturated regime

On the sweep above the warning fires at `c = 1` and `100` and stays quiet at `0.01` and `0.1`, which is the intended split.

Implementation is a no-op `TrainingSAE.training_convergence_warning()` hook, so other architectures are unaffected. It's skipped when resuming from a checkpoint, since travel is measured from initialization.

Two things I deliberately left alone: the `l0_coefficient=1.0` default (a behaviour change, and per the above no fixed value is right across setups), and the `5.0` in the `tanh` example, which isn't affected.

Happy to take this further — a target-L0 scheme is the option that would actually be robust to the scale-dependence — but that's a much larger change and a config-semantics decision, so I've kept this to documenting what the knob currently does.

Tests: 8 new, CPU-only, no model download. Existing suite passes.

### DIFF
--- a/docs/training_saes.md
+++ b/docs/training_saes.md
@@ -147,6 +147,32 @@ sparse_autoencoder = LanguageModelSAETrainingRunner(cfg).run()
 
 [JumpReLU SAEs](https://arxiv.org/abs/2407.14435) are a state-of-the-art SAE architecture. To train one, provide a `JumpReLUTrainingSAEConfig` to the `sae` field. JumpReLU SAEs use a sparsity penalty controlled by the `l0_coefficient` parameter. The `JumpReLUTrainingSAEConfig` also has parameters `jumprelu_bandwidth` and `jumprelu_init_threshold` which affect the learning of the thresholds.
 
+<!-- prettier-ignore-start -->
+!!! warning "Choosing `l0_coefficient` in the default `step` mode"
+
+    In the default `jumprelu_sparsity_loss_mode="step"`, the sparsity penalty reaches the
+    model only through the threshold. Adam caps the threshold's movement at roughly `lr`
+    per step regardless of gradient magnitude, so `l0_coefficient` does not set the
+    sparsity level directly — it sets the *direction* of the threshold's update, while
+    `lr * n_steps` sets how far the threshold can travel in total.
+
+    The practical consequence is that `l0_coefficient` **saturates**. Above some value the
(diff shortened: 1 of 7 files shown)


chanind:
This feels like overkill, #718 is merged so hopefully that should satisfy this

PR #724, opened 2026-08-14
fix: treat a BOS token id of 0 as set when concatenation is disabled

SuryanshSS1011:
# Summary:

`concat_and_batch_sequences` carries the batch-start token into the `disable_concat_sequences` path with a truthiness test, while every other token-id check in the file uses `is not None`:

```python
if begin_batch_token_id and not begin_sequence_token_id:
    begin_sequence_token_id = begin_batch_token_id
```

Token id `0` is the BOS id of the GPT-NeoX family, including Pythia, so it reads as absent. Two things follow: a requested BOS of `0` is never prepended, and an explicit `begin_sequence_token_id=0` is silently replaced by `begin_batch_token_id`.

This reaches training, beyond just pretokenization. `ActivationsStore._iterate_tokenized_sequences` passes `begin_batch_token_id=(bos_token_id if self.prepend_bos else None)` with `begin_sequence_token_id=None`, and `prepend_bos` defaults to `True`. So a SAE trained on Pythia with `disable_concat_sequences=True` gets activations with no BOS while the config records that BOS was prepended, with no warning.

```diff
-        if begin_batch_token_id and not begin_sequence_token_id:
+        if begin_batch_token_id is not None and begin_sequence_token_id is None:
```

The path is well covered, but every existing `disable_concat_sequences` test uses id `998` or `999`, so the one value that breaks the line was never exercised. I added two tests using `0`, one per failure mode. Both fail on `main`.

Existing runs that hit this now get the BOS they asked for, so their sequences shift by one token and drop the last token of each context, which is what the non-zero-BOS path already does. Nothing changes for gpt2 style tokenizers, and `disable_concat_sequences` defaults to `False`.

## Type of change

# Checklist:

### DIFF
--- a/sae_lens/tokenization_and_batching.py
+++ b/sae_lens/tokenization_and_batching.py
@@ -82,7 +82,7 @@ def concat_and_batch_sequences(
         max_batches: If not provided, the iterator will be run to completion.
     """
     if disable_concat_sequences:
-        if begin_batch_token_id and not begin_sequence_token_id:
+        if begin_batch_token_id is not None and begin_sequence_token_id is None:
             begin_sequence_token_id = begin_batch_token_id
         for sequence in tokens_iterator:
             if (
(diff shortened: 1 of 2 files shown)


chanind (approved):
Great catch! Thanks for this

PR #725, opened 2026-08-26
Fix SkipTranscoder parameter dtype and device

koriyoshi2041:
## Problem

`SkipTranscoder` initializes `W_skip` with a bare `torch.zeros` call. This leaves the parameter on CPU with the default dtype even when the transcoder is configured for another dtype or device. A float64 configuration then fails during the forward pass because the skip projection mixes float64 activations with a float32 parameter.

## Fix

Initialize `W_skip` with the transcoder's configured dtype and device, matching the other parameters. Add a regression that constructs a float64 transcoder, checks the parameter metadata, and runs a forward pass.

## Test

- `pytest tests/saes/test_transcoder.py` (52 passed)
- Ruff lint and format checks on the changed files
- `git diff --check`

## Risk

Low. The change only affects initial placement of the zero-initialized skip parameter; its shape and initial values are unchanged.

## Type of change

## Checklist

- [x] I have added tests that prove my fix is effective
- [x] New and existing focused tests pass locally

### DIFF
--- a/sae_lens/saes/transcoder.py
+++ b/sae_lens/saes/transcoder.py
@@ -228,7 +228,14 @@ class SkipTranscoder(Transcoder):
 
         # Initialize skip connection matrix
         # Shape: [d_out, d_in] to map from input to output dimension
-        self.W_skip = nn.Parameter(torch.zeros(self.cfg.d_out, self.cfg.d_in))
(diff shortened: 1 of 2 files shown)


chanind (approved):
Great catch!

PR #726, opened 2026-08-26
fix checkpoint ID generation with current W&B

koriyoshi2041:
## Summary

- replace the removed private `wandb.util.generate_id` helper with `uuid.uuid4`
- preserve the existing eight-character checkpoint ID format
- add a deterministic checkpoint-path regression

The current dependency range resolves W&B 0.28.2, where `wandb.util.generate_id` is no longer available. This currently breaks config construction across the test suite before unrelated tests can run.

## Validation

- `pytest -q tests/training/test_config.py` — 30 passed
- `ruff check sae_lens/config.py tests/training/test_config.py`
- `ruff format --check sae_lens/config.py tests/training/test_config.py`
- `git diff --check`

Risk is limited to the source of auto-generated run IDs; explicit `wandb_id` values and the checkpoint path shape are unchanged.

### DIFF
--- a/sae_lens/config.py
+++ b/sae_lens/config.py
@@ -1,9 +1,10 @@
 import json
 import math
+import uuid
 import warnings
 from dataclasses import asdict, dataclass, field
 from pathlib import Path
-from typing import TYPE_CHECKING, Any, Generic, Literal, TypeVar, cast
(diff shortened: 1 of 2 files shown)


chanind (approved):
Good call, thanks for fixing this!

PR #727, opened 2026-08-29
Validate empty Matryoshka level widths

koriyoshi2041:
## Problem

`MatryoshkaBatchTopKTrainingSAEConfig` defaults `matryoshka_widths` to an empty list, but `_validate_matryoshka_config` immediately indexes its final element. Omitting the hierarchy therefore raises an opaque `IndexError` instead of a configuration error.

## Fix

Reject an empty width list with a descriptive `ValueError` before inspecting the final level, and add a direct regression test using the real training config.

## Test

- `pytest tests/saes/test_matryoshka_batchtopk_sae.py -q` — 25 passed (with the separately isolated W&B compatibility prerequisite needed by the current resolved environment)
- `ruff check` and `ruff format --check` on both changed files
- `git diff --check`

## Risk

Low. Valid non-empty configurations are unchanged; only the exception type and message for an already-invalid empty hierarchy change.


chanind (approved):
Makes sense!

PR #732, opened 2026-09-03
feat: add quadratic sparcity loss from GemmaScope 2 for JumpReluSAE

Estomeee:
# Description

During the research, I needed to reproduce the learning process of JumpReluSAE from [GemmaScope2](https://storage.googleapis.com/deepmind-media/DeepMind.com/Blog/gemma-scope-2-helping-the-ai-safety-community-deepen-understanding-of-complex-language-model-behavior/Gemma_Scope_2_Technical_Paper.pdf). Their implementation uses some modifications of the loss function, which, as far as I can see, are not present in the current version of the library.

I implemented them for myself, but I thought it might be useful for other users working with JumpReluSAE.

I have added the ability to use the `quadratic l0 penalty` when training `JumpReluSAE.` The `Quadratic L0 penalty` allows you to set a target target number of active latents. In other words, the model aims not just to reduce the number of active latents, but to keep it close to a given target.
​
Made the minimum necessary changes to the documentation to reflect the new parameter.

Validation for target_l0 has also been implemented: target_l0 must be positive and cannot exceed the number of SAE latents. In terms of the validation variant and layout, I relied on the existing one for matryoshka SAE - [this](https://github.com/decoderesearch/SAELens/blob/main/sae_lens/saes/matryoshka_batchtopk_sae.py#L205).

I also added tests covering the new functionality. These include tests for valid and invalid target_l0 values, a test verifying that the quadratic L0 loss is zero when the actual L0 matches the target, and a test verifying the expected gradient direction when the actual L0 is above or below the target. The existing JumpReLU tests continue to pass as well.

## Type of change

# Checklist:

### Performance Check.

- [x] L0


koriyoshi2041 (commented):
The `2 / target_l0` normalization matches Eq. (5) in the Gemma Scope 2 technical paper, but the new config currently accepts `target_l0=0` and negative values. Zero raises `ZeroDivisionError` on the first quadratic-loss call, while a negative target makes the quadratic sparsity term negative, reversing the optimization incentive. The paper implicitly requires a strictly positive target because the target is both an expected active-latent count and the denominator.\n\nCould the config reject `target_l0 <= 0` before training starts, with focused tests for zero/negative targets plus a positive-target loss/gradient case? That would also keep malformed YAML/config input from failing deep inside the training loop.

chanind (approved):
Thanks for adding this!

PR #733, opened 2026-09-09
fix: preserve sequence starts at batch boundaries

kiteretsu903:
# Description

When a separator fills the last slot in a batch, the next document can lose its start token. With 128-token chunks, separate BOS/EOS markers, and no batch-start token, this can cause a final 127-token document to be dropped instead of forming a complete chunk.

Keep the start token pending until it has been emitted. Also skip empty inputs when concatenation is disabled, avoiding an `IndexError` with `context_size=1`. The existing multi-marker test now includes the previously missing token. No new dependencies.

## Type of change

# Checklist:

- [ ] I have not rewritten tests relating to key interfaces which would affect backward compatibility (two expected-output rows were corrected)

### Performance Check.

N/A


chanind (approved):
Makes sense, thanks for this fix!

PR #734, opened 2026-09-13
Validate positive Matryoshka widths

koriyoshi2041:
## Description

Reject zero and negative `matryoshka_widths` during configuration validation. These values currently pass validation but create empty or reverse prefix slices in the nested reconstruction path, so the configured level does not represent a valid non-empty prefix.

The existing strict-increase check, together with appending `d_sae`, already rejects widths above `d_sae`; this change closes the lower-bound gap.

No new dependencies.

## Type of change

## Checklist

- [x] I have added tests that prove my fix is effective
- [x] New and existing focused tests pass locally with my changes

## Validation

- `pytest tests/saes/test_matryoshka_batchtopk_sae.py -q` — 27 passed
- `ruff check` on both changed files — passed
- `ruff format --check` on both changed files — passed
- `git diff --check` — passed

This is configuration validation only and does not change training behavior for valid configurations, so the training performance metrics in the template are not applicable.


chanind (approved):
Fair enough!

PR #736, opened 2026-09-21
feat: add AbsTopK bidirectional SAE variant

smellslikeml:
# Description

Adds **AbsTopK**, a magnitude-ranked, sign-preserving SAE activation: it keeps the top-k pre-activations by absolute value with their sign preserved (no ReLU), so latents can fire in either direction (bidirectional features). It is registered alongside `TopK`/`BatchTopK` and reuses the existing SAE base classes and TopK machinery, so it plugs into training, eval, and loading with no new dependencies.

Paper: *AbsTopK: Rethinking Sparse Autoencoders For Bidirectional Features* (ICLR 2026, [arXiv:2510.00404](https://arxiv.org/abs/2510.00404)). Reimplemented from the MIT-licensed reference implementation ([GoXzascc/AbsTopK-SAE](https://github.com/GoXzascc/AbsTopK-SAE)) against SAELens's SAE base classes.

**One shared-file touch worth flagging:** `evals.py` computes L0 and feature-density with `!= 0` instead of `> 0`, so bidirectional AbsTopK's negative-but-active latents are counted. This is a no-op for ReLU-based SAEs (activations are ≥ 0) and is the enabling change for AbsTopK's sparsity metric — trivially reviewable, so I kept it in this PR rather than splitting it out.

Fixes # (n/a — new feature, no tracking issue)

## Type of change

# Checklist:

- [x] I have made corresponding changes to the documentation (`docs/custom_saes.md`)
- [x] I have added tests that prove my feature works (`tests/saes/test_abstopk_sae.py`)

### Performance Check.

This PR ships the AbsTopK operator, its registration, and tests; it does not change any existing training defaults, so no control/test dashboard is included (cf. #601 Matching Pursuit, merged without this section). Happy to add an L0 / CE / MSE sweep with control-vs-test dashboards if maintainers would like one before merge.

---

🤖 Drafted with Claude Code, human-reviewed. AI-assisted per this repo's Claude Code convention.


smellslikeml:
Thanks for the review — pushed b3182fe8 addressing it.

**1. Synthetic metrics omit negative activations (fixed).** Good catch — I'd applied the `!= 0` firing semantics in `evals.py` but not propagated it to `sae_lens/synthetic/evals.py`, and AbsTopK is reachable there via the registry through `SyntheticSAERunner`. `L0Calculator`, `DeadLatentsCalculator`, and `ClassificationMetricsCalculator` now test SAE-latent firing with `!= 0` (ground-truth acts stay `> 0`, since they're non-negative by construction). Added direct regression tests for `L0Calculator` / `DeadLatentsCalculator` on negative-active latents.

**2. `l1` aggregates signed activations (fixed).** Agreed — this was a gap in the same change: `l0` moved to `!= 0` but `l1` stayed a signed sum, which can cancel or go negative for bidirectional latents. It now sums magnitudes (`.abs().sum()`), and the existing all-negative-activation eval test asserts the L1 magnitude (would report `-15` instead of `15` under the old sum).

Both changes are no-ops for ReLU-based SAEs (activations ≥ 0).

**3. Batch-global selection semantics.** I believe this one is a false positive — selection is per-token: `torch.topk(x.abs(), k=self.k, dim=-1)` operates along the feature dimension independently for each row, and the tests assert exactly `k` survivors per token (largest by absolute value, sign preserved). Happy to look again if there's a specific path I'm missing.

**4. `d_in == 1` division by zero in the aux loss.** This mirrors the parent `calculate_topk_aux_loss` exactly (`k_aux = sae_in.shape[-1] // 2`, `scale = num_dead / k_aux`, no guard) — `d_in == 1` is a degenerate single-feature SAE that doesn't arise in practice, so I kept AbsTopK consistent with the existing TopK path rather than diverging. If you'd like a guard, I'm glad to add one to both in a small follow-up.

